PR #42, opened 2019-07-19
Corrections of UCTE Glsk convertor

amirakahya:

No

Bug fix : error during the conversion of UCTE glsk file

Generator id / load id created from ucte glsk file wich correspond to the node id

Correction of generator id and load id when creating the scalables from ucte glsk file

No


data/glsk-file/src/main/java/com/farao_community/farao/commons/data/glsk_file/actors/GlskDocumentScalableConverter.java
```diff
@@ -31,6 +31,8 @@
  */
 public class GlskDocumentScalableConverter {
 
+    private static final String TYPE_GLSK_FILE = "CIM";
```
murgeyseb: Prefer an Enum class for the GLSK types

PR #46, opened 2019-07-24
Share correction

amirakahya:

No

Bug fix : error when the value of share does not exist in the Glsk File

Error when the share does not exist

Set the value of share (shareFactor) at 100 when it does not exist in the Glsk File

No


data/glsk-file/src/main/java/com/farao_community/farao/commons/data/glsk_file/UcteGlskSeries.java
```diff
@@ -55,8 +55,11 @@ public UcteGlskSeries(Element element) {
...
         this.ucteBusinessType = ((Element) element.getElementsByTagName("BusinessType").item(0)).getAttribute("v");
         this.timeSeriesID = ((Element) element.getElementsByTagName("TimeSeriesIdentification").item(0)).getAttribute("v");
-        this.shareFactor = Double.parseDouble(((Element) element.getElementsByTagName("BusinessType").item(0)).getAttribute("share"));
-
+        if (!getShare(element).isEmpty()) {
+            this.shareFactor = Double.parseDouble(getShare(element));
+        } else {
+            this.shareFactor = 100;
+        }
         if (Objects.requireNonNull(element).getElementsByTagName("CountryGSK_Block").getLength() > 0) {
```
murgeyseb: Do not test the emptiness of the result, but prefer using `hasAttribute("share")` method of Element class

PR #97, opened 2019-12-05
Create a new RAO: LinearRangeActionRao

TheMaskedTurtle:

Structure of a new remedial action optimizer that deals only with linear remedial actions


ra-optimisation/linear-range-action-rao/src/main/java/com/farao_community/farao/linear_range_action_rao/LinearRangeActionRaoParameters.java
```diff
@@ -0,0 +1,36 @@
...
+        return "LinearRangeActionRaoParameters";
+    }
+
+    public SensitivityComputationParameters getSensitivityComputationParameters() {
+        return sensitivityComputationParameters;
+    }
+
+    public LinearRangeActionRaoParameters setSensitivityComputationParameters(SensitivityComputationParameters sensiParameters) {
+        this.sensitivityComputationParameters = Objects.requireNonNull(sensiParameters);
+        return this;
```
alexandrem-rte: why don't you do a classical setter returning void?

ra-optimisation/linear-range-action-rao/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+    <parent>
+        <artifactId>farao-ra-optimisation</artifactId>
+        <groupId>com.farao-community.farao</groupId>
+        <version>1.3.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>farao-linear-range-action-rao</artifactId>
+    <packaging>jar</packaging>
+    <name>Modular RAO on linear actions implementation</name>
```
baptsegu: I would say that the optimisation is linear, but not the actions. We approximate their impact with linear sensitivity coefficients, yet they are _a priori_ not linear.
Maybe : 
`Modular linear RAO implementation`

ra-optimisation/pom.xml
```diff
@@ -22,6 +22,7 @@
         <module>ra-optimisation-afs-local</module>
         <module>rao-api</module>
         <module>search-tree-rao</module>
+        <module>linear-range-action-rao</module>
```
baptsegu: alphabetical order

ra-optimisation/linear-range-action-rao/src/main/java/com/farao_community/farao/linear_range_action_rao/LinearRangeActionRaoParameters.java
TheMaskedTurtle: As it is copied from PowSyBl we'll wait to know if there is a reason

TheMaskedTurtle: Internet says that this pattern could be useful if you have a lot of parameters to set and you don't want to pass too many arguments in the constructor. For us I think we can remove it by now.

TheMaskedTurtle: After discussion with Sylvain, we will keep it for parameters classes consistency.

ra-optimisation/linear-range-action-rao/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+
+    <artifactId>farao-linear-range-action-rao</artifactId>
+    <packaging>jar</packaging>
+    <name>Modular linear RAO implementation</name>
+    <description>Implementation of a RAO that will optimize only linear range actions</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>farao-rao-api</artifactId>
```
pengbo-ruidian: dependency order

TheMaskedTurtle: ok

PR #98, opened 2019-12-06
Memory improvements on Closed Optimisation RAO

murgeyseb:

No

This PR allows to better control memory usage of the sensitivity pre-processor of closed optimisation RAO implementation.
Threshold had been created previously to avoid using sensitivity values too small in the optimisation problem. This PR adds a step in pre-processor to filter this values directly in the overall data map, saving a lot of memory.
Moreover, it adds a parameter to control the size of the thread pool used to compute sensitivities in parallel.

Currently all sensitivities are saved in the data map and the pre-processor uses all available threads of the machine

Sensitivities that may be filtered in the problem fillers are now filtered from the sensitivities pre processor, and the number of parallel threads used is provided by a dedicated parameter.

The default value for the number of parallel threads is now of 1. To allow parallel computation in the pre-processor, do not forget to add the parameter **number-of-parallel-threads** to any expected value to modify this default value.

JSON parameters names have been aligned with YAML ones.

**Other information**:
Includes small corrections on parameters initial checks


ra-optimisation/closed-optimisation-rao/src/main/java/com/farao_community/farao/closed_optimisation_rao/pre_processors/SensitivityPreProcessor.java
```diff
@@ -145,6 +149,21 @@ public void fillData(Network network, CracFile cracFile, ComputationManager comp
...
     }
 
+    private void initThreshold(Map<String, Object> data) {
+        if (data.containsKey(OPTIMISATION_CONSTANTS_DATA_NAME)) {
+            Map<String, Object> optimisationConstants = (Map<String, Object>) data.get(OPTIMISATION_CONSTANTS_DATA_NAME);
+            if (optimisationConstants.containsKey(RD_SENSITIVITY_SIGNIFICANCE_THRESHOLD_NAME)) {
+                redispatchingSensitivityThreshold = (Double) optimisationConstants.get(RD_SENSITIVITY_SIGNIFICANCE_THRESHOLD_NAME);
+            }
+            if (optimisationConstants.containsKey(PST_SENSITIVITY_SIGNIFICANCE_THRESHOLD_NAME)) {
+                redispatchingSensitivityThreshold = (Double) optimisationConstants.get(PST_SENSITIVITY_SIGNIFICANCE_THRESHOLD_NAME);
```
baptsegu: pstSensitivityThreshold here

murgeyseb: fixed

PR #142, opened 2020-02-06
Add in the config of the LinearRao a max number of iterations, and a parameter to skip the linear rao

baptsegu:

No

Add one parameter in the LinearRaoParameter, maxIterations. This parameter informs on the max number of iteration (one iteration = one sensi + one LP) that the LinearRao can do.
Add an other parameter, skipLinearRaoParameter (default false), which allows us to skip the linear rao computation if wanted (a simple security analysis is run instead)

This parameter can be loaded from .yml or .json config files. Under the name `max-number-of-iterations` 

maxIterations parameter is fixed (to 10) and hardcoded

maxIterations can be loaded from the config and can be chosen by the user.
The user is able to skip the linear rao computation if he sets the skip parameter to true.

No

**Other information**:


ra-optimisation/linear-rao/src/main/java/com/farao_community/farao/linear_rao/config/LinearRaoParameters.java
```diff
@@ -33,4 +38,13 @@ public LinearRaoParameters setSensitivityComputationParameters(SensitivityComput
         this.sensitivityComputationParameters = Objects.requireNonNull(sensiParameters);
         return this;
     }
+
+    public int getMaxIterations() {
+        return maxIterations;
+    }
+
+    public LinearRaoParameters setMaxIterations(int maxIterations) {
+        this.maxIterations = max(1, maxIterations);
```
phiedw: we could allow maxIterations to be 0 to skip linear rao computation entirely

distribution/src/main/resources/config.yml
```diff
@@ -57,3 +57,14 @@ local-app-file-system:
 mapdb-app-file-system:
     drive-name: Database
     db-file: $HOME/farao.db
+
+rao:
+    default: SearchTreeRao
+
+linear-rao-parameters:
+    max-number-of-iterations: 10
+    skip-linear-rao: false
```
baptsegu: The name is quite strange, as you run a linear RAO skipping the linear RAO.
(sorry :s)

Maybe : 
`security-analysis-without-rao` : false
`run-only-security-analysis` : false

ra-optimisation/linear-rao/src/main/java/com/farao_community/farao/linear_rao/LinearRao.java
```diff
@@ -56,16 +56,25 @@ public String getVersion() {
                                                        String variantId,
                                                        ComputationManager computationManager,
                                                        RaoParameters parameters) {
+        LinearRaoParameters linearRaoParameters = parameters.getExtensionByName("LinearRaoParameters");
+        if (linearRaoParameters == null) {
+            throw new FaraoException("rao parameters is missing linear rao parameters extension");
+        }
```
baptsegu: I would create a configuration quality check, as in the searchTreeRao.

phiedw: ok (only checks if linearRaoParameters exist for now)

distribution/src/main/resources/config.yml
phiedw: changed the name to 
security-analysis-without-rao : false

ra-optimisation/linear-rao/src/main/java/com/farao_community/farao/linear_rao/config/LinearRaoConfigurationUtil.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (c) 2019, RTE (http://www.rte-france.com)
```
baptsegu: 2020 :p

PR #149, opened 2020-02-12
Import single direction CNECs

phiedw:

No

allows the use of single direction CNECs

The getMinThreshold and getMaxThreshold functions act as if the CNEC threshold is in both directions.

The getMinThreshold and getMaxThreshold functions return different values depending on the direction of the threshold.

**Other information**:

### DIFF
--- a/data/crac-file/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
+++ b/data/crac-file/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
@@ -67,14 +67,22 @@ public abstract class AbstractFlowThreshold extends AbstractThreshold {
 
     @Override
     public Optional<Double> getMinThreshold(Unit requestedUnit) throws SynchronizationException {
-        // TODO : handle cases where direction != Direction.BOTH
-        return Optional.of(-convert(getAbsoluteMax(), unit, requestedUnit));
+        switch (direction) {
+            case OUT:
+                return Optional.of(Double.NEGATIVE_INFINITY);
+            default:
+                return Optional.of(-convert(getAbsoluteMax(), unit, requestedUnit));
+        }
     }
 
     @Override
     public Optional<Double> getMaxThreshold(Unit requestedUnit) throws SynchronizationException {
-        // TODO : handle cases where direction != Direction.BOTH
-        return Optional.of(convert(getAbsoluteMax(), unit, requestedUnit));
+        switch (direction) {
+            case IN:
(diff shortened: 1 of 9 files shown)


data/crac-file/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
```diff
@@ -67,14 +67,20 @@ public PhysicalParameter getPhysicalParameter() {
 
     @Override
     public Optional<Double> getMinThreshold(Unit requestedUnit) throws SynchronizationException {
-        // TODO : handle cases where direction != Direction.BOTH
-        return Optional.of(-convert(getAbsoluteMax(), unit, requestedUnit));
+        if (direction == Direction.OUT) {
```
baptsegu: Do we really want to keep Direction.IN and Direction.OUT ?
I would prefer (and @terriervik too), Direction.ORIGIN_TO_EXTREMITY and Direction.EXTREMITY_TO_ORIGIN, or (but we are less fans) Direction.DIRECT and Direction.INDIRECT

data/crac-file/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
```diff
@@ -67,14 +67,20 @@ public PhysicalParameter getPhysicalParameter() {
...
+            return Optional.of(Double.NEGATIVE_INFINITY);
```
baptsegu: We should return Optional.empty() -> It means there is no Threhold limits and in that case there is no constraint in the optimisation problem (and we do not build a dummy constraint Flow >= -infinity).

data/crac-file/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
phiedw: modified to Direction.DIRECT and Direction.OPPOSITE

data/crac-file/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
phiedw: switched to Optional.empty()
(modified a few Double.MIN_VALUE / Double.MAX_VALUE to Double.NEGATIVE_INFINITY / Double.POSITIVE_INFINITY)

PR #154, opened 2020-02-18
For windows compilation

pengbo-ruidian:


pengbo-ruidian:
line separartor in windows is \r\n, in linux is \n

PR #180, opened 2020-03-05
Refactor computeMargin

terriervik:

Refactoring

Currently the computeMargin(Network) method inside crac-api for cnec interface is possibly wrong because a cnec is associated to a state (which corresponds to a network state), so the margin of a cnec has to be computed on the corresponding network state. Here, as an argument, we have no guarantee that the network object will be the correct one associated to the good state.

Min and max thresholds will be directly accessible from the cnec so the only remaining thing is to implement a method which computes the margin directly from the flow (double value). The user can already do it by comparing the flow value and cnec min and max values but we can internalize this method inside the cnec to make it more readable.

### DIFF
--- a/data/crac/crac-api/src/main/java/com/farao_community/farao/data/crac_api/Cnec.java
+++ b/data/crac/crac-api/src/main/java/com/farao_community/farao/data/crac_api/Cnec.java
@@ -35,9 +35,9 @@ public interface Cnec extends Identifiable<Cnec>, Synchronizable {
      * positive, it means that the limits of the Threshold are respected. If it is negative,
      * it means that that a limit of the Threshold has been overcome.
      *
-     * margin = min(maxThreshold - actualValue, actualValue - minThreshold)
+     * margin = min(maxThreshold - actualFlow, actualFlow - minThreshold)
      */
-    double computeMargin(Network network, Unit requestedUnit);
+    double computeMargin(double actualFlow, Unit flowUnit);
 
     /**
      * A Threshold consists in monitoring a given physical value (FLOW, VOLTAGE
--- a/data/crac/crac-api/src/main/java/com/farao_community/farao/data/crac_api/Cnec.java
+++ b/data/crac/crac-api/src/main/java/com/farao_community/farao/data/crac_api/Cnec.java
@@ -35,9 +35,9 @@ public interface Cnec extends Identifiable<Cnec>, Synchronizable {
      * positive, it means that the limits of the Threshold are respected. If it is negative,
      * it means that that a limit of the Threshold has been overcome.
      *
-     * margin = min(maxThreshold - actualFlow, actualFlow - minThreshold)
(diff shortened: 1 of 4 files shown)


data/crac/crac-api/src/main/java/com/farao_community/farao/data/crac_api/Cnec.java
```diff
@@ -35,9 +35,9 @@
      * positive, it means that the limits of the Threshold are respected. If it is negative,
      * it means that that a limit of the Threshold has been overcome.
      *
-     * margin = min(maxThreshold - actualValue, actualValue - minThreshold)
+     * margin = min(maxThreshold - actualFlow, actualFlow - minThreshold)
      */
-    double computeMargin(Network network, Unit requestedUnit);
+    double computeMargin(double actualFlow, Unit flowUnit);
```
TheMaskedTurtle: It was value because CNEC is supposed to be working with voltage or angle monitoring...

ra-optimisation/linear-rao/src/main/java/com/farao_community/farao/linear_rao/LinearRao.java
```diff
@@ -224,11 +221,8 @@ private MonitoredBranchResult createMonitoredBranchResultAndUpdateLinearRaoResul
             throw new FaraoException(format("Cnec %s is not present in the linear RAO result. Bad behaviour.", cnec.getId()));
         }
 
-        double limitingThreshold;
-        double margin1 = cnec.getMaxThreshold(Unit.MEGAWATT).orElse(Double.POSITIVE_INFINITY) - postOptimFlow;
-        double margin2 = postOptimFlow - cnec.getMinThreshold(Unit.MEGAWATT).orElse(Double.NEGATIVE_INFINITY);
-        double marginPostOptim =  Math.min(margin1, margin2);
-        limitingThreshold = cnec.getMaxThreshold(Unit.MEGAWATT).orElse(-cnec.getMinThreshold(Unit.MEGAWATT).orElseThrow(FaraoException::new));
+        double marginPostOptim =  cnec.computeMargin(postOptimFlow, Unit.MEGAWATT);
+        double limitingThreshold = cnec.getMaxThreshold(Unit.MEGAWATT).orElse(-cnec.getMinThreshold(Unit.MEGAWATT).orElseThrow(FaraoException::new));
```
baptsegu: rename absoluteThreshold

terriervik: ok

data/crac/crac-api/src/main/java/com/farao_community/farao/data/crac_api/Cnec.java
terriervik: So actualValue and valueUnit?

PR #193, opened 2020-03-13
Add dc-mode and ac-to-dc-fallback parameters to SearchTreeRao

terriervik:

To be able to launch either AC or DC computations in Search Tree Rao from the configuration. In any case both sensitivities and simple load-flows computations must be performed in the chosen solution. Another parameter in the configuration would specify if the Linear Rao can use a fallback mode in case load-flows do not converge in AC. In this case the Rao must be able to manage this fallback situation.
These two parameters are added to the configuration (the DC and fallback behaviours are not developed yet).


util/src/main/java/com/farao_community/farao/util/SystematicSensitivityAnalysisService.java
```diff
@@ -36,12 +38,17 @@ private SystematicSensitivityAnalysisService() {
...
+                                                                  ComputationManager computationManager,
+                                                                  RaoParameters parameters) {
         String initialVariantId = network.getVariantManager().getWorkingVariantId();
 
         Map<State, SensitivityComputationResults> stateSensiMap = new HashMap<>();
         Map<Cnec, Double> cnecFlowMap = new HashMap<>();
 
+        SearchTreeRaoParameters searchTreeRaoParameters = parameters.getExtensionByName("SearchTreeRaoParameters");
+        boolean dcMode = searchTreeRaoParameters.isDcMode();
+        boolean acToDcFallback = searchTreeRaoParameters.isAcToDcFallback();
```
baptsegu: We are not sure that the searchTreeRaoParameters exist.
For instance, if we launch an "independent linear-rao", not within a search-tree-rao.

So concerning my first comment, I now think these parameters should be in RaoParameters (not SearchTreeRaoParameters).

baptsegu:
@TheMaskedTurtle what do you think about putting the parameters related to ac/dc in RaoParameter and not in SeachTreeRaoParameter

distribution/src/main/resources/config.yml
```diff
@@ -67,4 +67,6 @@ linear-rao-parameters:
 
 search-tree-rao-paramters:
     range-action-rao : LinearRao
+    dc-mode : false
+    ac-to-dc-fallback : false
```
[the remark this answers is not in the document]
TheMaskedTurtle: Yes indeed you're right !

util/src/main/java/com/farao_community/farao/util/SystematicSensitivityAnalysisService.java
TheMaskedTurtle: Do you mean in LinearRaoParameters ?

util/src/main/java/com/farao_community/farao/util/SystematicSensitivityAnalysisService.java
```diff
@@ -36,12 +38,17 @@ private SystematicSensitivityAnalysisService() {
 
     public static SystematicSensitivityAnalysisResult runAnalysis(Network network,
                                                                   Crac crac,
-                                                                  ComputationManager computationManager) {
+                                                                  ComputationManager computationManager,
+                                                                  RaoParameters parameters) {
```
TheMaskedTurtle: I think it's weird to pass RaoParameters here. Either you can create like a SensitivityComputationParameters or directly pass the two information as arguments (an enum and a boolean let's say). So that you don't make it too difficult considering that it will disappear after powsybl systematic sensi is integrated !

util/src/main/java/com/farao_community/farao/util/SystematicSensitivityAnalysisService.java
baptsegu: I don't like putting it in an extension of a RaoParameters because you never know if the extension is here or not. I prefer the solution of the RaoParameters, or a dedicated SystematicSensitivityAnalysisParameters.

util/src/main/java/com/farao_community/farao/util/SystematicSensitivityAnalysisService.java
baptsegu: SensitivityComputationParameters can be nice too. I agree :heart_decoration:

baptsegu: hmmm... or two new arguments as one of them is temporary

distribution/src/main/resources/config.yml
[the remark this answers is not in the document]
terriervik: Thanks for this nice specification ;)

ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/RaoParameters.java
```diff
@@ -50,6 +55,8 @@ public static RaoParameters load(PlatformConfig platformConfig) {
         Objects.requireNonNull(platformConfig);
 
         RaoParameters parameters = new RaoParameters();
+        //parameters.setDcMode(config.getBooleanProperty("dc-mode", DEFAULT_DC_MODE));
+        //parameters.setAcToDcFallback(config.getBooleanProperty("ac-to-dc-fallback", DEFAULT_AC_TO_DC_FALLBACK));
```
terriervik: @baptsegu @TheMaskedTurtle 
Do you know if I have to write such a thing here? (ie get the booleans from platformConfig)

TheMaskedTurtle: I would say yes otherwise you can never load the config. But I don't know how to test that properly.

util/src/test/java/com/farao_community/farao/util/SystematicSensitivityAnalysisServiceTest.java
```diff
@@ -52,6 +52,9 @@ public void setUp() {
         FileSystem fileSystem = Jimfs.newFileSystem(Configuration.unix());
         InMemoryPlatformConfig platformConfig = new InMemoryPlatformConfig(fileSystem);
 
+        /*raoParameters = Mockito.mock(RaoParameters.class);
+        Mockito.when(raoParameters.getExtensionByName("SearchTreeRaoParameters")).thenReturn(new SearchTreeRaoParameters());*/
+
```
TheMaskedTurtle: ?

util/src/main/java/com/farao_community/farao/util/SystematicSensitivityAnalysisService.java
```diff
@@ -33,7 +33,9 @@ private SystematicSensitivityAnalysisService() {
 
     public static SystematicSensitivityAnalysisResult runAnalysis(Network network,
                                                                   Crac crac,
-                                                                  ComputationManager computationManager) {
+                                                                  ComputationManager computationManager,
+                                                                  boolean dcMode,
+                                                                  boolean acToDcFallback) {
```
TheMaskedTurtle: The parameters are not used !!

terriervik: Sure, the aim of the dev was to import the parameters, not to implement the whole US...

util/src/test/java/com/farao_community/farao/util/SystematicSensitivityAnalysisServiceTest.java
terriervik: Yes, this was temporary.

distribution/src/main/resources/config.yml
```diff
@@ -60,12 +60,13 @@ mapdb-app-file-system:
...
     default: SearchTreeRao
+    dc-mode: false
+    ac-to-dc-fallback: false
 
 linear-rao-parameters:
     max-number-of-iterations: 10
     security-analysis-without-rao: false
 
-search-tree-rao-paramters:
+search-tree-rao-parameters:
```
terriervik: @baptsegu @TheMaskedTurtle 
Do you think that this config.yml is reliable after this? :fearful:  
By the way, I think that even my modifications are wrong (because I put them in rao, and not rao-parameters)...

baptsegu: Yeah, the new parameters should be in `search-tree-rao-parameters`.

Indeed, it means we never read this parameters when doing computation with our local config... we were using the default parameters (which are the same than the one below, that's why we didnt see the problem).

Thanks for the correction. If you want to make sure it works, you can change the local config and see in debug mode if the RaoParameters was changed accordingly.

util/src/main/java/com/farao_community/farao/util/SystematicSensitivityAnalysisService.java
baptsegu: :package:

distribution/src/main/resources/config.yml
baptsegu: my bad, it should be in `rao-parameters`, not` search-tree-rao-parameters`

ra-optimisation/linear-rao/src/test/resources/LinearRaoParameters.json
```diff
@@ -1,5 +1,7 @@
 {
   "version" : "1.0",
+  "dc-mode" : false,
+  "ac-to-dc-fallback" : false,
```
TheMaskedTurtle: is it useful ?

ra-optimisation/search-tree-rao/src/test/resources/SearchTreeRaoParameters.json
```diff
@@ -1,5 +1,7 @@
 {
   "version" : "1.0",
+  "dc-mode" : false,
+  "ac-to-dc-fallback" : false,
```
TheMaskedTurtle: Is it useful ?

ra-optimisation/linear-rao/src/test/resources/LinearRaoParameters.json
terriervik: This file is the expected result of the round trip test, and I guess we want to export these parameters, regardless if they are explicitly specified or not.

ra-optimisation/search-tree-rao/src/test/resources/SearchTreeRaoParameters.json
terriervik: same

PR #204, opened 2020-03-23
fixing export for contingencies

phiedw:

No

Bug fix

exporting a crac with contingencies writes the whole network elements in the contingencies making the json impossible to import

the contingencies now only contain the network element ids

No

**Other information**:


phiedw:
i think yeah

phiedw:
actually it doesn't since we've modified the crac export quite a bit

PR #221, opened 2020-04-08
Remove CRAC dependency to Sensitivity API

murgeyseb:

**Other information**:
Remove an unwanted dependency between CRAC-API and SENSITIVITY-API


util/src/main/java/com/farao_community/farao/util/SystematicSensitivityAnalysisResult.java
```diff
@@ -26,16 +31,45 @@ public SystematicSensitivityAnalysisResult(Map<State, SensitivityComputationResu
...
 
-    public Map<Cnec, Double> getCnecFlowMap() {
-        return cnecFlowMap;
+    public Optional<Double> getIntensity(Cnec cnec) {
+        return Optional.ofNullable(cnecIntensityMap.get(cnec));
     }
 
-    public Map<Cnec, Double> getCnecIntensityMap() {
-        return cnecIntensityMap;
+    public Optional<Double> getSensitivity(Cnec cnec, State state, RangeAction rangeAction) {
```
TheMaskedTurtle, line 42: Actually I think you can remove the state as argument here and take directly the cnec state. Because this could lead to inconsistencies.

murgeyseb: Done

util/src/main/java/com/farao_community/farao/util/CracFactorsProvider.java
```diff
@@ -0,0 +1,116 @@
...
+ */
+public class CracFactorsProvider implements SensitivityFactorsProvider {
+    private final Crac crac;
+
+    public CracFactorsProvider(Crac crac) {
+        this.crac = crac;
+    }
+
+    @Override
+    public List<SensitivityFactor> getFactors(Network network) {
```
baptsegu, line 42: How the notion of State is handled by this method ? If I need to build sensitivity factors in order to compute the sensi on different network variants (N and N-1s) :

* the SensitivityFactors are not dependent from the State (they just use the information of network ID and names), and can therefore be used whatever the State for which we want to run the SensitivityAnalysis ?

* or, do I need to execute this getFactors() method on different Network, each network being updated with the topology of the N/N-1s State.

baptsegu: Maybe we can add something in the javadoc on this.

util/src/main/java/com/farao_community/farao/util/CracFactorsProvider.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private SensitivityFunction cnecToSensitivityFunction(Network network, Cnec cnec) {
+        String id = cnec.getId();
+        String name = cnec.getName();
+        String branchId = cnec.getNetworkElement().getId();
+        Identifiable networkIdentifiable = network.getIdentifiable(branchId);
+        if (networkIdentifiable instanceof Branch) {
+            return new BranchFlow(id, name, branchId);
+        } else {
+            throw new FaraoException("Unable to create sensitivity funtion for " + id);
```
baptsegu, line 82: function instead of funtion

util/src/main/java/com/farao_community/farao/util/CracFactorsProvider.java
murgeyseb: If you look at how the API of sensi is designed for systematic sensi analysis in powsybl-core snapshot, ContingenciesProvider and FactorsProvider are independant. We suppose that we want to observe same things in N and N-k, whichis not always correct.
I think the best is to have something that can work in any configuration (on any state).

And for the variant, the answer is still in the sensitivity computation API. In the run method, you provide the desired working variant. But as befoore, this will change a lot when upgrading powsybl version and using systematic computation....

util/src/main/java/com/farao_community/farao/util/CracFactorsProvider.java
murgeyseb: fixed

PR #230, opened 2020-04-24
Corrected minimim and maximum PST angles when range is given in CRAC

pet-mit:

Bug in PstWithRange

For a PST with positive deltaU, when a range is given in the CRAC, its intersection with the range given in the network file behaves badly : upper and lower bounds of the intersection are inverted

The lower and upper bounds of the intersection are correctly computed
Tests have been added


murgeyseb:
I am a bit puzzled by this fix... I really think this mapping problems with network should be done in client module, not in the CRAC itself.

pet-mit:
Shouldn't the getMinValue (resp. getMaxValue) method always return the minimum (resp. maximum) absolute angle of the PST? If that's not an issue, then it would be easier to do this in the client code (LinearRao)

PR #237, opened 2020-05-12
Single out loopflow from flowbased

pengbo-ruidian:

prepare for dev in regression test


pengbo-ruidian:
> I think you can remove dependencies on flowbased-computation in linear rao and search tree rao poms. And I was just wondering if it was possible to move GlskProvider into data/glsk-file so that your loopflow computation is also not dependent on flowbased-computation ?

Nice!
Could you put the idea of GlskProvider refactoring somewhere in backlog ? or in a US, but preferably not under US 7.* :)

PR #242, opened 2020-05-14
Add pst penalty cost parameter

TheMaskedTurtle:

Enhancement of a parameter that was hardcoded to put it in configuration file

Currently pst taps movement has a penalty cost in the objective function which is hardcoded.

The new behavior is to load this penalty cost in the configuration. Default value is 0.5.

No breaking changes because a default value is set.

### DIFF
--- a/ra-optimisation/linear-rao/src/main/java/com/farao_community/farao/linear_rao/config/LinearRaoParameters.java
+++ b/ra-optimisation/linear-rao/src/main/java/com/farao_community/farao/linear_rao/config/LinearRaoParameters.java
@@ -23,6 +23,7 @@ public class LinearRaoParameters extends AbstractExtension<RaoParameters> {
     static final int DEFAULT_MAX_NUMBER_OF_ITERATIONS = 10;
     static final boolean DEFAULT_SECURITY_ANALYSIS_WITHOUT_RAO = false;
     static final double DEFAULT_PST_SENSITIVITY_THRESHOLD = 0.0;
+    static final double DEFAULT_PST_PENALTY_COST = 0.5;
 
     private SensitivityComputationParameters sensitivityComputationParameters = new SensitivityComputationParameters();
     private SensitivityComputationParameters fallbackSensiParameters = null;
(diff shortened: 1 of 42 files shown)


ra-optimisation/linear-rao/src/test/java/com/farao_community/farao/linear_rao/config/LinearRaoParametersTest.java
```diff
@@ -61,4 +61,10 @@ public void setPstSensitivityThreshold() {
         raoParameters.setPstSensitivityThreshold(5.0);
         assertEquals(5.0, raoParameters.getPstSensitivityThreshold(), DOUBLE_TOLERANCE);
     }
+
+    @Test
+    public void setPstPenaltyCost() {
+        raoParameters.setPstPenaltyCost(0.5);
```
murgeyseb, line 67: Try the set method with a value different than default one

ra-optimisation/linear-rao/src/main/java/com/farao_community/farao/linear_rao/config/LinearRaoParameters.java
```diff
@@ -23,13 +23,15 @@
     static final int DEFAULT_MAX_NUMBER_OF_ITERATIONS = 10;
     static final boolean DEFAULT_SECURITY_ANALYSIS_WITHOUT_RAO = false;
     static final double DEFAULT_PST_SENSITIVITY_THRESHOLD = 0.0;
+    static final double DEFAULT_PST_PENALTY_COST = 0.5;
```
murgeyseb, line 26: Why did you change default value (which was 1.0 previously) ?

TheMaskedTurtle: it is specified to be set at 0.01. I previously put 0.5 to not break cucumber tests. Now I respect the specification so I will have another PR in cucumber tests

ra-optimisation/linear-rao/src/test/java/com/farao_community/farao/linear_rao/config/LinearRaoParametersTest.java
TheMaskedTurtle: now default value is different :smile:

PR #247, opened 2020-05-15
Fix root evaluation not switching back to initial variant

phiedw:

No

Bug fix

After evaluating the root we do not switch back to the initial variant. This means we start the next iterations with the optimized PSTs etc

We clone the initial variant and switch back to it after evaluating the root.

**Other information**:

### DIFF
--- a/ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/process/search_tree/Tree.java
+++ b/ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/process/search_tree/Tree.java
@@ -18,7 +18,6 @@ import com.farao_community.farao.search_tree_rao.config.SearchTreeRaoParameters;
 import com.farao_community.farao.util.FaraoNetworkPool;
 import com.powsybl.iidm.network.Network;
 
-import java.util.Collections;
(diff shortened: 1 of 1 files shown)


ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/process/search_tree/Tree.java
```diff
@@ -53,7 +55,11 @@ private Tree() {
         SearchTreeRaoParameters searchTreeRaoParameters = parameters.getExtensionByName("SearchTreeRaoParameters");
 
         Leaf rootLeaf = new Leaf();
-        rootLeaf.evaluate(network, crac, referenceNetworkVariant, parameters);
+        String initialNetworkVariant = network.getVariantManager().getWorkingVariantId();
+        String newNetworkVariant = RandomizedString.getRandomizedString(Collections.singleton(initialNetworkVariant));
```
murgeyseb, line 59: replace singletonList by network.getVariantManager().getVariantIds()

phiedw: :+1:

PR #249, opened 2020-05-18
Improvements after real-size data tests

TheMaskedTurtle:

- Add logs
- Add a parameter to set parallel computing on leaves (by default 1 thread)
- Convert TreeMap to HashMap in systematic sensitivity computation results
- Compute sensitivities on not duplicated network elements and not on all cnecs (that can duplicate network elements)
- Fix a bug on relative and absolute margin for tree next iteration. Previously it was comparing with the optimal leaf but the optimal leaf was set during the loop so it had to be compared with the initial optimal leaf value


ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/process/search_tree/Tree.java
```diff
@@ -169,4 +191,48 @@ static RaoResult buildOutput(Leaf rootLeaf, Leaf optimalLeaf) {
...
+
+    private static String generateLeafResults(Leaf leaf, Crac crac) {
+        String rangeActionResults = crac.getRangeActions()
+            .stream()
+            .map(rangeAction -> format("%s: %d",
+                rangeAction.getName(),
+                ((PstRangeResult) rangeAction.getExtension(RangeActionResultExtension.class)
+                    .getVariant(leaf.getRaoResult().getPostOptimVariantId()))
+                    .getTap(crac.getPreventiveState().getId())))
+            .collect(Collectors.joining(", "));
```
baptsegu, line 211: why not also logs the network actions results ?

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/process/search_tree/Tree.java
```diff
@@ -169,4 +191,48 @@ static RaoResult buildOutput(Leaf rootLeaf, Leaf optimalLeaf) {
         raoResult.setPostOptimVariantId(optimalLeaf.getRaoResult().getPostOptimVariantId());
         return raoResult;
     }
+
+    private static void logOptimalLeaf(Leaf leaf, Crac crac) {
+        LOGGER.info(format("Optimal leaf - %s", generateLeafResults(leaf, crac)));
```
baptsegu, line 196: I don't know if logging is time consuming, but it might be quite a big log line on "real-size" data with 100 remedial actions. Maybe put a INFO version with the number of RA activated, and a DEBUG version with the details for all RA.

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/process/search_tree/Tree.java
```diff
@@ -169,4 +191,48 @@ static RaoResult buildOutput(Leaf rootLeaf, Leaf optimalLeaf) {
...
+    }
+
+    private static void logLeafResults(Leaf leaf, Crac crac) {
+        LOGGER.info(generateLeafResults(leaf, crac));
```
baptsegu, line 200: same as above.

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/process/search_tree/Tree.java
```diff
@@ -94,22 +106,32 @@ private Tree() {
                 }
                 currentLeaf.deletePreOptimResultVariant(crac);
             }
+
+            if (hasImproved) {
+                logOptimalLeaf(optimalLeaf, crac);
+            } else {
```
baptsegu, line 112: Isn't this redundant with the `logOptimalLeaf(optimalLeaf, crac);` (line 95) of the next iteration ?

TheMaskedTurtle: Capture d’écran de 2020-05-20 14-09-09
The idea was to have at one glance, the current optimal leaf, all the generated leaves and the chosen one. So at each search tree iteration you can really understand what happened.

TheMaskedTurtle: So yes it is redundant but I thought it could make the logs clearer

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/process/search_tree/Tree.java
TheMaskedTurtle: It is done just below these lines.
`return` format("%s: minimum margin = %f (%s)",
            leaf.isRoot() ? "root leaf" : leaf.getNetworkActions().stream().map(NetworkAction::getName).collect(Collectors.joining(", ")),
            -leaf.getCost(crac),
            rangeActionResults);`

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/process/search_tree/Tree.java
TheMaskedTurtle: so when you're in debug mode you will get the "minimal" line AND the "complete" line following each other ?...

TheMaskedTurtle: I separated NetworkAction and RangeAction information. RangeAction information will only appear at debug level but NetworkAction information is still at info level. But we have to keep in mind that the list of the combined network actions  size is the tree depth and I think for now it won't be higher than a few (like 5). I think we can consider the issue later if thus list becomes very long

baptsegu: Yes, I was more thinking of the RangeAction

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/process/search_tree/Tree.java
baptsegu: ok, it's just one line anyway

PR #251, opened 2020-05-20
New parameters in linear rao config

baptsegu:
- [X] Docs have been added / updated (for bug fixes / features)

Two new parameters are added in the LinearRao config, which can be imported from a classif yml config file or a JSON LinearRaoParameter file.
The parameters are : 

* `objective-function`, which can take two possibles values, `MAX_MIN_MARGIN_IN_MEGAWATT` and `MAX_MIN_MARGIN_IN_AMPERE`. The default behavior currently in Farao is to maximum the min margin in megawatt. The other possible value - whose behavior will be later implemented - is to maximize instead the min margin in ampere. 
* `sensitivity-fallback-overcost` (double), the overcost associated to a situation which run with the fallback sensitivity configuration (typically DC) but not with the default sensitivity configuration (typically AC). In order to prioritize leaves with a slighty worst margins but whose default sensitivity configuration converged, this fallback overcost can be set at a strictly positive value to penalise the leaves which didn't converge in default configuration.

**Other information**:
Those parameters are not handle today in the optimisation.


ra-optimisation/linear-rao/src/main/java/com/farao_community/farao/linear_rao/config/LinearRaoParameters.java
```diff
@@ -74,6 +84,15 @@ public LinearRaoParameters setMaxIterations(int maxIterations) {
         return this;
     }
 
+    public ObjectiveFunction getObjectiveFunction() {
+        return objectiveFunction;
+    }
+
+    public LinearRaoParameters setObjectiveFunction(ObjectiveFunction objectiveFunction) {
```
terriervik, line 91: Do you really need the setter to return something? (and not just void?)

baptsegu: This pattern is used for all the configs of PowSyBl.

We were also surprised about that, but we discussed about it with @TheMaskedTurtle to PowSyBl experts which said it was useful in some cases to chain the setters... After this discussion we decided to keep the PowSyBl logics, even if it is not useful for us it gives more "uniformity". :p

PR #252, opened 2020-05-25
Loopflow PTDF approximation

pengbo-ruidian:

In current loopflow computation, ptdf is re-calculated each time a loopflow is called. In this PR a new configuration parameter is added as "loopflow approximation". If this parameter is set to "true", then loopflow computation will use previous saved ptdf without a new sensi.


ra-optimisation/linear-rao/src/main/java/com/farao_community/farao/linear_rao/optimisation/fillers/MaxLoopFlowFiller.java
```diff
@@ -48,9 +48,14 @@ public void update(LinearRaoData linearRaoData, LinearRaoProblem linearRaoProble
...
-        LoopFlowComputation loopFlowComputation = new LoopFlowComputation(linearRaoData.getCrac()); //todo optimization => do not recalculate loopflowShifts, use the one in the CracResult
-        Map<Cnec, Double> loopFlowShifts = loopFlowComputation.buildZeroBalanceFlowShift(linearRaoData.getNetwork()); //todo: validate assumption that "loopflowShift" does not change "much"
+    private void buildMaxLoopFlowConstraint(LinearRaoData linearRaoData, LinearRaoProblem linearRaoProblem, LinearRaoParameters linearRaoParameters) {
+        Map<Cnec, Double> loopFlowShifts;
+        LoopFlowComputation loopFlowComputation = new LoopFlowComputation(linearRaoData.getCrac());
+        if (!Objects.isNull(linearRaoParameters.getExtendable()) && linearRaoParameters.getExtendable().isLoopflowApproximation()) {
+            loopFlowShifts = loopFlowComputation.buildLoopflowShiftsApproximation(linearRaoData.getCrac());
+        } else {
+            loopFlowShifts = loopFlowComputation.buildZeroBalanceFlowShift(linearRaoData.getNetwork());
+        }
```
baptsegu, line 58: I'm not sure of what I say, but should'nt we always use the previously stored shifts here, as : 

* if we have` "loopflow-approximation" : false`, we want to be quick and not rerun a loopflow computation
* if we have` "loopflow-approximation" : true`, the loopflow have already been computed on the same network in the SensitivityAnalysisEngine (either during the first sensi analysis of the LinearRao, either during the sensi of the previous iteration of the LinearRao)

What do you think ? As I understand it we compute 2 times the loopflow per iteration of the LinearRao (one here in the filler, and one in the sensiEngine) and we are doing 2 times the same calculation.

ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/RaoParameters.java
```diff
@@ -74,15 +75,33 @@ private void readExtensions(PlatformConfig platformConfig) {
...
 
+    /**
+     *  loopflow approximation means using previous calculated ptdf and net position values to compute loopflow
+     *  ptdf is calculated by a sensitivity analysis, net position is derived from current network
+     *  if "loopflowApproximation" is set to "false", then each loopflow computation do a sensi for ptdf;
+     *  if "loopflowApproximation" is set to "true", loopflow computation tries to use previous saved ptdf and netposition.
+     *  note: Loopflow = reference flow - ptdf * net position
+     */
+    private boolean loopflowApproximation = DEFAULT_LOOPFLOW_APPROXIMATION;
+
```
baptsegu, line 90: I like it.

Not for this PR, but maybe for later... Do you think if would be relevant to have some kind of "in between" approximation ? which would be : 
* we compute the loopflows 1 time per LinearRao, at the beginning of the LinearRao
* except for this first time, inside the LinearRao, we also use the previously calculated shifts

As the LinearRao is called several time during a search-tree, I would be more accurate than the current approximation, but quicker than the accurate mode.

We will have to find the right tradeoff between accuracy and computation times for the loopflow, having those different modes will enable us to do those testing and advise the users for the best configuration.

ra-optimisation/linear-rao/src/main/java/com/farao_community/farao/linear_rao/optimisation/fillers/MaxLoopFlowFiller.java
pengbo-ruidian: "loopflow-approximation = false" means that we compute the exact loopflow by re-running a sensi for PTDF; " = true" means that we use previous calculated PTDF * netposition value (as mentioned in US7.8).

ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/RaoParameters.java
pengbo-ruidian: Sure, we can add another parameter to do a inital ptdf for each LinearRao, this means we need to save the ptdf * netpostion value with variant. 

You can have a check already with the current parameter to test if loopflow value changes a lot between a exact ptdf and an approximated ptdf.

ra-optimisation/linear-rao/src/main/java/com/farao_community/farao/linear_rao/optimisation/fillers/MaxLoopFlowFiller.java
baptsegu: Could you please try to always go in the if part of this if/else and see if it changes the results on some cases where  `"loopflow-approximation" : false`. I am pretty sure that the shifts have already been calculated on the same network here and that we do not need to go for the "accurate" loop-flow computation.

baptsegu: It might not be the case for the first iteration with the current design though... let's wait for the next meeting on this topic.

PR #272, opened 2020-06-10
management of FRM during computation

alexandrem-rte:
Add an attribute to AbstractFlowThreshold which contains its Cnec's frm in MEGAWATT
A setMargin method has been added at AbstractThreshold level
Adaptation of the methods getMinThreshold and getMaxThreshold of AbstractFlowThreshold to take into account this new attribute
Adaptation of the method copy of AbsoluteFlowThreshold and RelativeFlowThreshold

**Other information**:

### DIFF
--- a/data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
+++ b/data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
@@ -137,8 +137,12 @@ public abstract class AbstractFlowThreshold extends AbstractThreshold {
         this.side = side;
     }
 
-    public void setMargin(double frmInMW) {
-        this.frmInMW = frmInMW;
+    public void setMargin(double frmInMW, Unit unit) {
+        if (unit != Unit.MEGAWATT) {
+            throw new FaraoException("Unable to handle another margin than the FRM in Megawatts");
+        } else {
+            this.frmInMW = frmInMW;
+        }
     }
 
     /**
--- a/data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
+++ b/data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
@@ -51,7 +51,7 @@ public abstract class AbstractFlowThreshold extends AbstractThreshold {
     /**
      * Flow reliability margin of the Cnec associated to this threshold, in MEGAWATT
      */
-    private double frmInMW;
+    protected double frmInMW;
 
     protected boolean isSynchronized;
 
@@ -204,7 +204,9 @@ public abstract class AbstractFlowThreshold extends AbstractThreshold {
             result = threshold.networkElement == null;
         }
         return result && unit.equals(threshold.unit)
-            && side.equals(threshold.side) && direction.equals(threshold.direction);
+            && side.equals(threshold.side)
+                && direction.equals(threshold.direction)
+                && frmInMW == (threshold.frmInMW);
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
```diff
@@ -72,7 +77,9 @@ public PhysicalParameter getPhysicalParameter() {
         if (direction == Direction.DIRECT) {
             return Optional.empty();
         } else { // Direction.OPPOSITE and Direction.BOTH
-            return Optional.of(-convert(getAbsoluteMax(), unit, requestedUnit));
+            return Optional.of(
+                    convert(frmInMW, Unit.MEGAWATT, requestedUnit)
+                    - convert(getAbsoluteMax(), unit, requestedUnit));
```
pet-mit, line 82: add some unit tests to check that the returned value is ok in different cases ? 
- FRM = 0
- FRM != 0
- different units

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
```diff
@@ -81,7 +88,9 @@ public PhysicalParameter getPhysicalParameter() {
         if (direction == Direction.OPPOSITE) {
             return Optional.empty();
         } else { // Direction.DIRECT and Direction.BOTH
-            return Optional.of(convert(getAbsoluteMax(), unit, requestedUnit));
+            return Optional.of(
+                    convert(getAbsoluteMax(), unit, requestedUnit)
+                    - convert(frmInMW, Unit.MEGAWATT, requestedUnit));
```
pet-mit, line 93: same

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractThreshold.java
```diff
@@ -104,4 +104,6 @@ public boolean isSynchronized() {
 
     @Override
     public abstract int hashCode();
+
+    public abstract void setMargin(double margin);
```
pet-mit, line 108: Maybe we should add a unit argument to this function ? calls would be less ambiguous

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
```diff
@@ -48,6 +48,11 @@
      */
     protected double voltageLevel;
 
+    /**
+     * Flow reliability margin of the Cnec associated to this threshold, in MEGAWATT
+     */
+    protected double frmInMW;
```
baptsegu, line 54: It seems quite tricky with the voltageLevel which is in this class. But isn't there a way to avoid the duplication of the frm (which is here and in the SimpleCnec) ?

The clean way would be to put the voltageLevel in the Cnec, but this is quite a huge refactoring. And I don't think we want to do this now.

Anyway, if we keep this parameter I think we could add a in the JSON import/export a case with a Cnec which as FRM != 0.

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
alexandrem-rte: this is partially done, because testing a threshold conversion cannot be done in a unit test (it recquires a synchronisation with a network)

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbstractFlowThreshold.java
alexandrem-rte: same

PR #288, opened 2020-07-02
Handle MNECs in optimization

pet-mit:

Handles Monitored elements (MNECs) in optimization

MNECs are treated like CNECs

MNECS should only be monitored as to not have their margin reduced; we do not need to optimize their margin

### DIFF
--- a/ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/RaoUtil.java
+++ b/ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/RaoUtil.java
@@ -50,15 +50,14 @@ public final class RaoUtil {
         if (raoParameters.getObjectiveFunction().equals(RaoParameters.ObjectiveFunction.MAX_MIN_MARGIN_IN_AMPERE)
             || raoParameters.getObjectiveFunction().equals(RaoParameters.ObjectiveFunction.MAX_MIN_MARGIN_IN_MEGAWATT)) {
             fillers.add(new MaxMinMarginFiller(raoParameters.getObjectiveFunction().getUnit(), raoParameters.getPstPenaltyCost()));
-            fillers.add(new MnecFiller(raoParameters.getMnecAcceptableMarginDiminution(), raoParameters.getMnecViolationCost(), raoParameters.getMnecConstraintAdjustmentCoefficient()));
+            fillers.add(new MnecFiller(raoParameters.getObjectiveFunction().getUnit(), raoParameters.getMnecAcceptableMarginDiminution(), raoParameters.getMnecViolationCost(), raoParameters.getMnecConstraintAdjustmentCoefficient()));
         }
         if (raoParameters.isRaoWithLoopFlowLimitation()) {
             fillers.add(createMaxLoopFlowFiller(raoParameters));
             return new IteratingLinearOptimizerWithLoopFlows(fillers, systematicSensitivityComputation,
-                createCostEvaluator(raoParameters), createVirtualCostEvaluator(raoParameters), createIteratingLoopFlowsParameters(raoParameters));
+                createObjectiveFunction(raoParameters), createIteratingLoopFlowsParameters(raoParameters));
         } else {
-            return new IteratingLinearOptimizer(fillers, systematicSensitivityComputation, createCostEvaluator(raoParameters),
-                    createVirtualCostEvaluator(raoParameters), createIteratingParameters(raoParameters));
+            return new IteratingLinearOptimizer(fillers, systematicSensitivityComputation, createObjectiveFunction(raoParameters), createIteratingParameters(raoParameters));
         }
     }
 
@@ -76,23 +75,13 @@ public final class RaoUtil {
             raoParameters.getFallbackOverCost(), raoParameters.isLoopFlowApproximation(), raoParameters.getLoopFlowViolationCost());
     }
 
-    public static CostEvaluator createCostEvaluator(RaoParameters raoParameters) {
+    public static ObjectiveFunctionEvaluator createObjectiveFunction(RaoParameters raoParameters) {
         if (raoParameters.getObjectiveFunction().equals(RaoParameters.ObjectiveFunction.MAX_MIN_MARGIN_IN_AMPERE)) {
-            return new MinMarginEvaluator(Unit.AMPERE);
+            return new MinMarginObjectiveFunction(Unit.AMPERE, raoParameters.getMnecAcceptableMarginDiminution(), raoParameters.getMnecViolationCost());
         } else if (raoParameters.getObjectiveFunction().equals(RaoParameters.ObjectiveFunction.MAX_MIN_MARGIN_IN_MEGAWATT)) {
-            return new MinMarginEvaluator(Unit.MEGAWATT);
+            return new MinMarginObjectiveFunction(Unit.MEGAWATT, raoParameters.getMnecAcceptableMarginDiminution(), raoParameters.getMnecViolationCost());
         } else {
             throw new NotImplementedException("Not implemented objective function");
-        }
-    }
-
-    public static CostEvaluator createVirtualCostEvaluator(RaoParameters raoParameters) {
-        if (raoParameters.getObjectiveFunction().equals(RaoParameters.ObjectiveFunction.MAX_MIN_MARGIN_IN_AMPERE)) {
-            return new MnecViolationCostEvaluator(Unit.AMPERE, raoParameters.getMnecAcceptableMarginDiminution(), raoParameters.getMnecViolationCost());
-        } else if (raoParameters.getObjectiveFunction().equals(RaoParameters.ObjectiveFunction.MAX_MIN_MARGIN_IN_MEGAWATT)) {
-            return new MnecViolationCostEvaluator(Unit.MEGAWATT, raoParameters.getMnecAcceptableMarginDiminution(), raoParameters.getMnecViolationCost());
-        } else {
-            throw new NotImplementedException("Not implemented virtual objective function");
         }
     }
 }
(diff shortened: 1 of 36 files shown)


ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/RaoUtil.java
```diff
@@ -53,14 +50,15 @@ public static IteratingLinearOptimizer createLinearOptimizer(RaoParameters raoPa
         if (raoParameters.getObjectiveFunction().equals(RaoParameters.ObjectiveFunction.MAX_MIN_MARGIN_IN_AMPERE)
             || raoParameters.getObjectiveFunction().equals(RaoParameters.ObjectiveFunction.MAX_MIN_MARGIN_IN_MEGAWATT)) {
             fillers.add(new MaxMinMarginFiller(raoParameters.getObjectiveFunction().getUnit(), raoParameters.getPstPenaltyCost()));
+            fillers.add(new MnecFiller(raoParameters.getMnecAcceptableMarginDiminution(), raoParameters.getMnecViolationCost(), raoParameters.getMnecConstraintAdjustmentCoefficient()));
```
alexandrem-rte, line 53: FYI this is this part of the code that I think we must refactor in another pull request. Here I would put RaoParameters as the unique argument of the MnecFiller constructor. However this is not possible because RaoParameters cannot be imported in rao_commons package otherwise it would cause a cyclic dependency. That is why I think we must move RaoParameters and some methods of RaoUtil into the rao_commons package. This must be discussed further and can wait a little bit.

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/MnecViolationCostEvaluator.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public double getCost(RaoData raoData) {
+        return getViolationCost(raoData);
+    }
+
+    private double getViolationCost(RaoData raoData) {
+        if (Math.abs(mnecViolationCost) < 1e-10) {
+            return 0;
```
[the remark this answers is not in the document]
pet-mit: Only to skip computations in case we don't care about MNECs violating their soft constraint (ie if mnecViolationCost = 0)

ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/RaoUtil.java
pet-mit: An advantage of the current implementations remains that the fillers only access the parameters they need in RaoParameters. This prevents them from doing things that are not in their scope.
Of course we will discuss this further

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/MnecViolationCostEvaluator.java
```diff
@@ -0,0 +1,92 @@
...
+import static com.farao_community.farao.commons.Unit.AMPERE;
+import static com.farao_community.farao.commons.Unit.MEGAWATT;
+
+/**
+ * An evaluator that computes the virtual cost resulting from the violation of
+ * the MNEC minimum margin soft constraint
+ *
+ * @author Peter Mitri {@literal <peter.mitri at rte-france.com>}
+ */
+public class MnecViolationCostEvaluator implements CostEvaluator {
```
baptsegu, line 26: In link with my comment above.

I guess there is need for :
- "elementary blocks" of CostEvaluator (CostEvaluator, with an interface similar to this one).
- "combined" CostEvaluator, which give the full vision of an objective function.

baptsegu: For instance the MinMarginCombinedEvaluator (MinMarginFunction) would rely on several "elementary blocks" such as : MinMarginCostEvaluator, MnecViolationCostEvaluator, LoopFlowViolationCostEvaluator, FallBackCostEvaluator, ...

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/MnecViolationCostEvaluator.java
```diff
@@ -0,0 +1,92 @@
...
+                double initialMargin = cnec.computeMargin(initialFlow, unit);
+                double newFlow = (unit == MEGAWATT) ? raoData.getSystematicSensitivityAnalysisResult().getReferenceFlow(cnec) :
+                        raoData.getSystematicSensitivityAnalysisResult().getReferenceIntensity(cnec);
+                double newMargin = cnec.computeMargin(newFlow, unit);
+                double convertedAcceptableMarginDiminution = mnecAcceptableMarginDiminution / getUnitConversionCoefficient(cnec, raoData);
+                totalMnecMarginViolation += Math.max(0, Math.min(0, initialMargin - convertedAcceptableMarginDiminution) - newMargin);
+            }
+        }
+        if (mnecsSkipped) {
+            LOGGER.warn("Some MNECs were skipped during violation cost evaluation because their sensitivity results were not available.");
```
baptsegu, line 69: i would be more specific "initial flow results were not available"

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/MnecViolationCostEvaluator.java
```diff
@@ -0,0 +1,92 @@
...
+                    // Sensitivity results are not available, skip cnec
+                    // (happens on search tree rao rootleaf evaluation)
+                    mnecsSkipped = true;
+                    continue;
+                }
+                double initialMargin = cnec.computeMargin(initialFlow, unit);
+                double newFlow = (unit == MEGAWATT) ? raoData.getSystematicSensitivityAnalysisResult().getReferenceFlow(cnec) :
+                        raoData.getSystematicSensitivityAnalysisResult().getReferenceIntensity(cnec);
+                double newMargin = cnec.computeMargin(newFlow, unit);
+                double convertedAcceptableMarginDiminution = mnecAcceptableMarginDiminution / getUnitConversionCoefficient(cnec, raoData);
```
baptsegu, line 64: If i understood well, it means that the mnecAcceptableMarginDiminution should be in MW, right ?

Maybe you could add the units in comment of the RaoParameter, otherwhise i'm afraid we forgot them quickly ;)

I guess (but not sure)
mnecAcceptableMarginDiminution : MW
adjustmentCoefficient : MW
mnecViolationCost : MW or A depending on the unit of the objective function

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/MnecViolationCostEvaluator.java
```diff
@@ -0,0 +1,92 @@
...
+                totalMnecMarginViolation += Math.max(0, Math.min(0, initialMargin - convertedAcceptableMarginDiminution) - newMargin);
+            }
+        }
+        if (mnecsSkipped) {
+            LOGGER.warn("Some MNECs were skipped during violation cost evaluation because their sensitivity results were not available.");
+        }
+        return mnecViolationCost * totalMnecMarginViolation;
```
baptsegu, line 71: I'm not sure this is consistent with what is in the LP.

If I'm not wrong, here the violation cost is proportional to the margin in A or MW (depending on the unit in the objective function). In the LP it is proportional to the margin in MW.

Maybe it would be easier to say : the mnecViolationCost is always in {unit of the objective function} / MW. That is to say we always multiply it with MW value.
It might also avoid some problem later when the objective function will be neither MW neither A (for instance with the relative margin)

pet-mit: mnecViolationCost has no real unit : it is either in "equivalent A virtual cost per A constraint violation", or "equivalent MW virtual cost per MW constraint violation"
so i think it is possible to use its value both in the evaluator and in the LP without worrying about units

pet-mit: I agree but we would be less coherent with the functional part of the objective if we change this (MinMarginEvaluator)

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/MnecViolationCostEvaluator.java
pet-mit: will do

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/MnecViolationCostEvaluator.java
pet-mit: will do

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/MnecViolationCostEvaluator.java
pet-mit: done

ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/RaoUtil.java
```diff
@@ -53,14 +50,15 @@ public static IteratingLinearOptimizer createLinearOptimizer(RaoParameters raoPa
...
         }
         if (raoParameters.isRaoWithLoopFlowLimitation()) {
             fillers.add(createMaxLoopFlowFiller(raoParameters));
             return new IteratingLinearOptimizerWithLoopFlows(fillers, systematicSensitivityComputation,
-                createCostEvaluator(raoParameters), createIteratingLoopFlowsParameters(raoParameters));
+                createCostEvaluator(raoParameters), createVirtualCostEvaluator(raoParameters), createIteratingLoopFlowsParameters(raoParameters));
```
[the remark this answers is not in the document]
pet-mit: done

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/MnecViolationCostEvaluator.java
pet-mit: will change the objective function in LP to use AMPERES when needed

baptsegu: perfect

PR #293, opened 2020-07-10
Intermediate crac refactoring

alexandrem-rte:

**Other information**:


data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/threshold/AbsoluteFlowThreshold.java
```diff
@@ -48,11 +53,24 @@ private void setMaxValue(double maxValue) {
...
+        }
+        return OptionalDouble.empty();
+    }
+
+    @JsonIgnore
+    public OptionalDouble getMaxValue() {
+        if (getDirection().equals(Direction.BOTH) || getDirection().equals(Direction.DIRECT)) {
+            return OptionalDouble.of(maxValue);
+        }
+        return OptionalDouble.empty();
```
baptsegu, line 69: I would have keep it even more simple with simple getters which returns the raw initial value, without converting it depending on the Threshold.

It's kindof my bad because I told you about a getMin() and a getMax() method. But i guess that the getMax() is sufficient for flow thresholds.

The getDirection() and getSide() could be put in the API as well. Then we have all the raw values for a cnec threshold.

PR #301, opened 2020-07-29
Loopflow threshold with unit

baptsegu:

- enable to define the loop-flow threshold with different units (A, MW, or %Imax) -> previously, the loop-flow threshold were implicitly defined in MW

- handle the unit conversion of loop-flow threshold in the RAO

- the upper bound for the loop-flows in the RAO is now : loop-flow-threshold **- frm** -> previously the frm was not considered

- handle JSON import/export of CnecLoopFlowExtensions -> previsouly, no JSON import/export

Yes, modification of the constuctor of CnecLoopFlowExtension


data/crac/crac-loopflow-extension/src/main/java/com/farao_community/farao/data/crac_loopflow_extension/CnecLoopFlowExtension.java
```diff
@@ -6,59 +6,73 @@
...
-    private double loopflowShift; //ptdf * net position
+    // INPUTS
+    private double inputThreshold;
+    private Unit inputThresholdUnit;
+    /*
+     - if the unit is PERCENT_IMAX, the input flow threshold should be between 0 and 100
+     - in the loop-flow threshold, PERCENT_IMAX is considered as a percentage of the Cnec
+       threshold (retrieved from the Crac, without considering the frm), and NOT as a percentage
+        of the branch current limit (retrieved from the Network)
+     */
```
TheMaskedTurtle, line 33: I think it's clearer to put the comment above the variable

data/crac/crac-loopflow-extension/src/main/java/com/farao_community/farao/data/crac_loopflow_extension/CnecLoopFlowExtension.java
```diff
@@ -77,4 +91,76 @@ public boolean hasLoopflowShift() {
...
+    private double getNominalVoltage(Network network) {
+        Branch branch = network.getBranch(this.getExtendable().getNetworkElement().getId());
+        if (branch == null) {
+            throw new FaraoException(String.format("Cnec with id %s was not found in the network", this.getExtendable().getId()));
+        }
+        return branch.getTerminal1().getVoltageLevel().getNominalV();
+    }
+
+    private double getCnecThresholdInA(Network network) {
+        return Math.min(getExtendable().getMaxThreshold(Unit.AMPERE).orElse(Double.POSITIVE_INFINITY),
```
TheMaskedTurtle, line 163: - Isn't it -FRM and not +FRM ? (actually I'm not even sure you should consider FRM here as long as you want to compute Imax)

- What if we have infinite values here ? Isn't going to fail ? Shouldn't we ensure that the threshold final value is not infinite ?

data/crac/crac-loopflow-extension/src/main/java/com/farao_community/farao/data/crac_loopflow_extension/JsonCnecLoopFlowExtensionSerializer.java
```diff
@@ -0,0 +1,71 @@
...
+        jsonGenerator.writeStartObject();
+        jsonGenerator.writeNumberField("inputThreshold", cnecLoopFlowExtension.getInputThreshold());
+        jsonGenerator.writeObjectField("inputThresholdUnit", cnecLoopFlowExtension.getInputThresholdUnit().name());
+        jsonGenerator.writeEndObject();
+    }
+
+    @Override
+    public CnecLoopFlowExtension deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
+        double inputThreshold = 100.0;
+        Unit unit = Unit.PERCENT_IMAX;
```
TheMaskedTurtle, line 39: I think it's better to put these default values in the constructor of the extension instead

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoDataManager.java
```diff
@@ -128,13 +130,15 @@ public void fillCracResultsWithSensis(double cost, double overCost) {
...
         raoData.getCrac().getCnecs(raoData.getCrac().getPreventiveState()).forEach(cnec -> {
             CnecLoopFlowExtension cnecLoopFlowExtension = cnec.getExtension(CnecLoopFlowExtension.class);
+
             if (!Objects.isNull(cnecLoopFlowExtension)) {
+                double loopFlowThreshold = Math.abs(cnecLoopFlowExtension.getInputThreshold(Unit.MEGAWATT, network));
                 double initialLoopFlow = Math.abs(loopFlows.get(cnec.getId()));
-                double loopFlowThreshold = Math.abs(cnecLoopFlowExtension.getInputLoopFlow());
-                cnecLoopFlowExtension.setLoopFlowConstraint(Math.max(initialLoopFlow, loopFlowThreshold));
+
+                cnecLoopFlowExtension.setLoopFlowConstraintInMW(Math.max(initialLoopFlow, loopFlowThreshold - cnec.getFrm()));
```
TheMaskedTurtle, line 141: Once again I think the FRM management is unclear here

baptsegu: What we want for the loop flow constraint is : 
(1 - minRamFactor) * Fmax - frm

where Fmax is the flow limit on the Cnec (as defined in the CBCORA file, without considering the frm)

The problem is that in our crac-api, the getMaxThreshold() method return (Fmax - frm), and not Fmax.

So basically, we are doing : 
(1 - minRamFactor) * (getMaxThreshold + FRM) - FRM

This is indeed unclear, but the crac-api as it is now is quite unclear also, and I refrain myself from refactoring anything :yum:

TheMaskedTurtle: Ok I get it ! And where is the (1 - minRamFactor) ?

baptsegu: sorry, (1 - minRamFactor) would be (inputThreshold) defined with Unit.PERCENT_IMAX.

the (1-minRamFactor) computation is handled in the CBCORA importer

data/crac/crac-loopflow-extension/src/main/java/com/farao_community/farao/data/crac_loopflow_extension/CnecLoopFlowExtension.java
baptsegu: I am adding a check in the maxLoopFlowFiller to ensure that we do not build a constraint if the threshold is inifinite

The threshold would only be infinite if the Cnec has no threshold.

data/crac/crac-loopflow-extension/src/main/java/com/farao_community/farao/data/crac_loopflow_extension/JsonCnecLoopFlowExtensionSerializer.java
baptsegu: I don't really like the idea of a default constructor because it would require to define method `setInputThreshold()` and `setInputThresholdUnit()`, while those values should be final as we do not want to change them during the RAO.

Actually i could be more demanding here and impose that the two values are given in the JSON file (and so remove the default value). Anyway if the JSON crac has been provided by the JsonCracExporter, it will always contains the two values.

data/crac/crac-loopflow-extension/src/main/java/com/farao_community/farao/data/crac_loopflow_extension/CnecLoopFlowExtension.java
baptsegu: oki

data/crac/crac-loopflow-extension/src/main/java/com/farao_community/farao/data/crac_loopflow_extension/JsonCnecLoopFlowExtensionSerializer.java
TheMaskedTurtle: Ok then just define final static variables for these values ! ;)

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoDataManager.java
TheMaskedTurtle: Alright ! All clear !

PR #304, opened 2020-08-03
Create network aliases creation class inside FARAO to adapt the need

TheMaskedTurtle:

The problem is that we miss a lot of network element at cleaning step because some of the UCTE nodes are described in the CRAC with the wrong bar bus bar number compared with what is present in the network. So this PR enables to identify a line without taking the busbar number into account. So whatever is the topology inside the substation a branch can still be identified.

A lot of elements are removed from the CRAC because we cannot map it between the CRAC and the network.

Elements describing a branch even with different busbar numbers are taken into account


ra-optimisation/rao-commons/src/test/java/com/farao_community/farao/rao_commons/RaoUcteAliasesCreationTest.java
```diff
@@ -0,0 +1,29 @@
...
+ * @author Joris Mancini {@literal <joris.mancini at rte-france.com>}
+ */
+public class RaoUcteAliasesCreationTest {
+
+    @Test
+    public void createAliases() {
+        Network network = Importers.loadNetwork("TestCase12Nodes_withXnodes.uct", getClass().getResourceAsStream("/TestCase12Nodes_withXnodes.uct"));
+        RaoUcteAliasesCreation.createAliases(network);
+
+        assertSame(network.getBranch("DDE2AA   DDE3AA   1"), network.getBranch("DDE2AA   DDE3AA   ROT"));
```
baptsegu, line 26: In the UCTE description, we have : 
` 1st character: UCTE country code   ///////
   2nd- 6th character:    short description of the geographical spot   /////////
   7th character: voltage level     //////////
   8th character: letter or figure for differentiating bus bars (optional)`

Shouldn't we be tolerant only on the 8th character of the node, and not on the last character of the node ?

ra-optimisation/rao-commons/src/test/java/com/farao_community/farao/rao_commons/RaoUcteAliasesCreationTest.java
```diff
@@ -0,0 +1,29 @@
...
+        assertSame(network.getBranch("DDE3AA   X_DEFR1  1"), network.getBranch("FFR2AA   X_DEFR1  1"));
```
baptsegu, line 27: could you add a test with a branch with Xnode and ElementName.

And also check that the getBranch with the "correct ID" (all 8 characters like in the UCT) is the same than the getBranch where the last characters of the ids are modified.

ra-optimisation/rao-commons/src/test/java/com/farao_community/farao/rao_commons/RaoUcteAliasesCreationTest.java
TheMaskedTurtle: I think you're right... I tested with 12 nodes cases, and we did wrong inside it... the 8th caracter is always a white space this why I added the trim() method.
I'll fix that !

ra-optimisation/rao-commons/src/test/java/com/farao_community/farao/rao_commons/RaoUcteAliasesCreationTest.java
TheMaskedTurtle: Sure !

PR #316, opened 2020-08-27
decrease memory usage when building SystematicSensitivityComputationResult

baptsegu:


pet-mit:
Obsolete (sensitivity computation has been refactored)

PR #331, opened 2020-09-23
modified checkSensi to use provider cnecs, added loadflowProvider

phiedw:

**Other information**:

### DIFF
--- a/sensitivity-computation/src/main/java/com/farao_community/farao/sensitivity_computation/LoadflowProvider.java
+++ b/sensitivity-computation/src/main/java/com/farao_community/farao/sensitivity_computation/LoadflowProvider.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.farao_community.farao.sensitivity_computation;
+
+import com.farao_community.farao.data.crac_api.NetworkElement;
+import com.powsybl.iidm.network.*;
+import com.powsybl.sensitivity.SensitivityFactor;
+import com.powsybl.sensitivity.SensitivityFunction;
+import com.powsybl.sensitivity.SensitivityVariable;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * @author Philippe Edwards {@literal <philippe.edwards at rte-france.com>}
+ */
+public class LoadflowProvider extends RangeActionSensitivityProvider {
+
+    @Override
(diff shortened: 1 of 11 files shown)


sensitivity-computation/src/main/java/com/farao_community/farao/sensitivity_computation/LoadflowProvider.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * To run a systematic sensitivity analysis and evaluate the flows in all states at once,
+ * hades requires sensitivities. We therefore extend RangeActionSensitivityProvider to use
+ * some of its conversion methods, and use a random PST from the network to create "dummy"
+ * sensitivities for each studied cnec.
+ *
+ * @author Philippe Edwards {@literal <philippe.edwards at rte-france.com>}
+ */
+public class LoadflowProvider extends RangeActionSensitivityProvider {
```
baptsegu, line 26: shouldn't it be the opposite ? the RangeActionSensitivityProvider which extends the LoadFlowProvider.

As the List of RangeAction of the RangeActionSensitivityProvider is not used in this class.

baptsegu: and functionnaly it seems strange to have access here to some methods of the RangeActionSensitivityProvider, such as : addSensitivityFactors(Set<RangeAction> rangeActions, Set<Cnec> cnecs)

sensitivity-computation/src/main/java/com/farao_community/farao/sensitivity_computation/RangeActionSensitivityProvider.java
```diff
@@ -68,7 +68,7 @@ private boolean willBeKeptInSensi(Generator gen) {
         return gen.getTerminal().isConnected() && gen.getTerminal().getBusBreakerView().getBus().isInMainSynchronousComponent();
     }
 
-    private SensitivityVariable defaultSensitivityVariable(Network network) {
+    protected SensitivityVariable defaultSensitivityVariable(Network network) {
```
baptsegu, line 71: could be then moved to the LoadFlowProvider is this class now extends LoadFlowProvider

baptsegu (changes requested):

sensitivity-computation/src/main/java/com/farao_community/farao/sensitivity_computation/LoadflowProvider.java
phiedw: You're right, i switched the classes around

sensitivity-computation/src/main/java/com/farao_community/farao/sensitivity_computation/RangeActionSensitivityProvider.java
phiedw: yeah I had to move a few methods

baptsegu (approved):

PR #335, opened 2020-09-29
get flow on branch which are not in the primary connex component

baptsegu:

No

cnec.getP(network) returns a **FaraoException** if the flow value in the network is NaN and the branch is not in the primary connex component of the network.
---> can cause a crash of the RAO if loop-flows are monitored on such a branch.

cnec.getP(network) returns **0.0** if the flow value in the network is NaN and the branch is not in the primary connex component of the network.


phiedw (dismissed):
seems good to me

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/SimpleCnec.java
```diff
@@ -132,6 +132,10 @@ private boolean isCnecDisconnected(Network network) {
         return !branch.getTerminal1().isConnected() || !branch.getTerminal2().isConnected();
     }
 
+    private boolean isCnecInMainComponent(Network network) {
+        return getTerminal(network).getBusView().getBus().isInMainConnectedComponent();
```
murgeyseb, line 136: I would prefer:
```java
return getTerminal(network).getBusView().getBus().isInMainSynchronousComponent();
```
because usually loadflow and sensitivity computation engines calculate islands based on synchronous components

murgeyseb (changes requested):

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/SimpleCnec.java
baptsegu: corrected

phiedw (approved):

PR #344, opened 2020-10-16
Pst tap convention 

pascolithe:

No

Feature

The PST setpoint was only in old convention Convergence (Starts at one)

Both conventions can now be used for PST setpoint (Starts at one and Centered on zero)


data/crac/crac-impl/src/test/java/com/farao_community/farao/data/crac_impl/remedial_action/network_action/PstSetpointTest.java
```diff
@@ -44,26 +47,52 @@ public void basicMethods() {
...
-            50);
+                "out_of_bound",
+                new NetworkElement(networkElementId),
+                50,
+                RangeDefinition.STARTS_AT_ONE);
+        try {
+            pstSetpoint.apply(network);
+            fail();
+        } catch (FaraoException e) {
+            assertEquals(String.format("Tap value 50 not in the range of high and low tap positions [-16,16] of the phase tap changer %s steps", networkElementId), e.getMessage());
```
TheMaskedTurtle, line 70: he message can be disturbing here, if you set your setpoint with STARTS_AT_ONE, you expect you PST to be [1, 33] for example. So I think you can adapt the message whether it starts at one or not.

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/remedial_action/network_action/PstSetpoint.java
```diff
@@ -63,15 +82,20 @@ public PstSetpoint(String id, NetworkElement networkElement, double setpoint) {
...
         PhaseTapChanger phaseTapChanger = network.getTwoWindingsTransformer(networkElement.getId()).getPhaseTapChanger();
-        if (phaseTapChanger.getHighTapPosition() - phaseTapChanger.getLowTapPosition() + 1 >= setpoint && setpoint >= 1) {
+
+        if (rangeDefinition.equals(CENTERED_ON_ZERO)
+                && setpoint > phaseTapChanger.getLowTapPosition() && setpoint < phaseTapChanger.getHighTapPosition()) {
+            phaseTapChanger.setTapPosition((int) setpoint);
+        } else if (rangeDefinition.equals(STARTS_AT_ONE)
+                && (phaseTapChanger.getHighTapPosition() - phaseTapChanger.getLowTapPosition() + 1 >= setpoint && setpoint >= 1)) {
             phaseTapChanger.setTapPosition(phaseTapChanger.getLowTapPosition() + (int) setpoint - 1);
         } else {
```
TheMaskedTurtle, line 92: Here we assume, that the PowSyBl phase tap changer is always centered on zero which I'm not sure is always correct

TheMaskedTurtle (changes requested):

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/remedial_action/network_action/PstSetpoint.java
pet-mit: is there a way to know which convention is used in the network ?

data/crac/crac-impl/src/test/java/com/farao_community/farao/data/crac_impl/json/CracImportExportTest.java
```diff
@@ -74,58 +75,60 @@ public void cracTest() {
...
+                "RTE",
+                Collections.singletonList(new OnConstraint(UsageMethod.AVAILABLE, preventiveState, preventiveCnec1)),
+                Collections.singletonList(new Range(-3, 3, RangeType.RELATIVE_DYNAMIC, CENTERED_ON_ZERO)),
+                Stream.of(simpleCrac.getNetworkElement("pst"), simpleCrac.addNetworkElement("pst2")).collect(Collectors.toSet())
         ));
 
         simpleCrac.setNetworkDate(new DateTime(2020, 5, 14, 11, 35));
 
         Crac crac = roundTrip(simpleCrac, SimpleCrac.class);
+
```
pet-mit, line 131: can you test that the RangeType is correctly exported & imported on some range actions here ?

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/remedial_action/network_action/PstSetpoint.java
```diff
@@ -29,30 +33,45 @@
 @JsonSerialize(using = PstSetPointSerializer.class)
 public final class PstSetpoint extends AbstractSetpointElementaryNetworkAction {
 
+    protected RangeDefinition rangeDefinition;
```
pet-mit, line 36: why not make this private (+getter) ?

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/remedial_action/network_action/PstSetpoint.java
pet-mit: is there a way to know which convention is used in the network ?

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/remedial_action/network_action/PstSetpoint.java
pascolithe: Done. I did it that way because I saw it done like this in the parent class AbstractSetpointElementaryNetworkAction

data/crac/crac-impl/src/test/java/com/farao_community/farao/data/crac_impl/json/CracImportExportTest.java
pascolithe: Done

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/remedial_action/network_action/PstSetpoint.java
pascolithe: Its done

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/remedial_action/network_action/PstSetpoint.java
```diff
@@ -85,14 +85,15 @@ public void apply(Network network) {
 
         if (rangeDefinition.equals(CENTERED_ON_ZERO)
                 && setpoint > phaseTapChanger.getLowTapPosition() && setpoint < phaseTapChanger.getHighTapPosition()) {
```
TheMaskedTurtle, line 87: Actually the conditions also assume that PST convention is centered on zero. You could define a "normalizedSetPoint" on which you apply the transformation according to range definition and then you can check independently if it's between low tap and high tap.
Another possibility is to catch ValidationException that would be thrown by PowSyBl, the problem is that this exception is thrown in implementation but is not declared in the interface, so it's not very clean. 
Last possibility is to assume that check on tap validity is on powsybl side so we don't do any catch or check on this.

I think solution 1 is the one that changes the least the current logic so maybe less controversial, but maybe not the ideal one not sure.

TheMaskedTurtle (changes requested):

data/crac/crac-impl/src/main/java/com/farao_community/farao/data/crac_impl/remedial_action/network_action/PstSetpoint.java
pascolithe: I implemented your first solution

TheMaskedTurtle (approved):

PR #356, opened 2020-11-02
Improve CRAC usage performance

murgeyseb:

No

Performance improvements

Currently, elements are stored in Sets in Simple Crac implementation, whereas research is often done by ID.
This implies poor performances in CRAC object usage in many algorithm (and importers).

Elements are now stored in maps, with ID as keys, improving research by ID.

Yes, it now returns unmodifiable sets instead of standard sets, reason why new methods **remove** have been added in CRAC interface to allow CRAC filtering through RaoInputHelper


TheMaskedTurtle (approved):
Maybe we should wait to confirm that it does improve performance and does not increase memory usage or is it fine ?

PR #359, opened 2020-11-04
Do not provide intensity sensitivity factors in dc

baptsegu:
**Rely on RC of powsybl-rte-core, and is therefore pending for next powsybl-rte-core release**

Improvement of the sensitivity-analysis API.
Enhancement of performance in DC, with less RAM occupation.

The SensitivityProvider provides factor in intensity, even if the computation is made in DC, resulting in : 
- useless big map of Double.NaN in the SystematicSensitivityAnalysisResult (i.e. useless RAM occupation)
- tons of error logs in PowSyBl, which filters the factor intensity factor in DC.

In DC, the intensity factors are not anymore provided by the sensitivity providers.

Yes, changes in the sensitivity-analysis API :

- The LoadFlowProvider,  RangeActionSensitivityProvider and PtdfSensitivityProvider now takes explictely in argument a set of Unit, defining wether the factor should be provided in A, MW or both.

- The sensitivityAnalysis run() method does not take anymore a Unit in argument. This argument does not really make sense anymore as a sensitivity analysis can be run with a MultipleSensitivityProvider with a mix of factors in MW (e.g. PTDF) and factors in A (e.g. RA sensi for an objective function in A). Moreover, this argument was previously only used in the checkSensiResult() method. Instead, a more intuitive check is now made at the beginning of the RAO, checking that the objective function unit is compatible with the load flow configuration.


sensitivity-analysis/src/main/java/com/farao_community/farao/sensitivity_analysis/LoadflowProvider.java
```diff
@@ -90,39 +106,32 @@ private String  faraoExceptionSensitivityString(SensitivityFunction function, Se
         return "Unable to create sensitivity factor for function of type " + function.getClass().getTypeName() + " and variable of type " + variable.getClass().getTypeName();
     }
 
-    protected List<SensitivityFunction> cnecToSensitivityFunctions(Network network, NetworkElement networkElement) {
+    List<SensitivityFunction> getSensitivityFunctions(Network network) {
+        Set<NetworkElement> networkElements = new HashSet<>();
+        cnecs.forEach(cnec -> networkElements.add(cnec.getNetworkElement()));
+        return networkElements.stream()
+            .map(networkElement -> cnecToSensitivityFunctions(network, networkElement))
+            .flatMap(Collection::stream)
```
TheMaskedTurtle, line 114: What is flatMap for ?

sensitivity-analysis/src/main/java/com/farao_community/farao/sensitivity_analysis/RangeActionSensitivityProvider.java
```diff
@@ -24,17 +22,29 @@
...
 
-    public void addSensitivityFactors(Set<RangeAction> rangeActions, Set<Cnec> cnecs) {
-        this.rangeActions.addAll(rangeActions);
-        super.addCnecs(cnecs);
+    @Override
+    public List<SensitivityFactor> getFactors(Network network) {
+        List<SensitivityFactor> factors = new ArrayList<>();
+        List<SensitivityVariable> sensitivityVariables = rangeActions.stream()
+            .map(ra -> rangeActionToSensitivityVariables(network, ra))
+            .flatMap(Collection::stream)
```
TheMaskedTurtle, line 38: same

TheMaskedTurtle (approved):

PR #365, opened 2020-11-09
Added a new "in-between" level of loopflow approximations

pet-mit:

New feature

The loopflows have two levels of approximation:
1- no approximation (loopflow-approximation=false): PTDFs are computed after each topological and PST change
2- heavy approximation (loopflow-approximation=true): PTDFs are computed only once at the start of the RAO
Moreover, in case 1, sensitivities were unnecessarily computed once more in the linear problem filler

We now have 3 levels of LF approximations:
1- no approximation (loopflow-approximation=UPDATE_PTDF_WITH_TOPO_AND_PST): same as above, but optimized to avoid computing sensitivities inside the linear problem filler
2- "in-between" approximation (loopflow-approximation=UPDATE_PTDF_WITH_TOPO): PTDFs are recomputed only after topological actions change. This is an exact method for DC computations, as for DC PSTs do not change PTDFs.
3- heavy approximation (loopflow-approximation=FIXED_PTDF): same as old case 2

Instead of setting loopflow-approximation=false/true in their configuration file, users should now use one of the three values: UPDATE_PTDF_WITH_TOPO_AND_PST / UPDATE_PTDF_WITH_TOPO / FIXED_PTDF

### DIFF
--- a/ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoUtil.java
+++ b/ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoUtil.java
@@ -87,7 +87,7 @@ public final class RaoUtil {
         }
     }
 
-    public static SystematicSensitivityInterface createSystematicSensitivityInterface(RaoParameters raoParameters, RaoData raoData, Boolean withPtdfSensitivities) {
+    public static SystematicSensitivityInterface createSystematicSensitivityInterface(RaoParameters raoParameters, RaoData raoData, boolean withPtdfSensitivities) {
 
         SystematicSensitivityInterface.SystematicSensitivityInterfaceBuilder builder = SystematicSensitivityInterface
             .builder()
@@ -96,11 +96,7 @@ public final class RaoUtil {
             .withRangeActionSensitivities(raoData.getAvailableRangeActions(), raoData.getCnecs());
 
         if (raoParameters.isRaoWithLoopFlowLimitation() && withPtdfSensitivities) {
-
             builder.withPtdfSensitivities(raoData.getGlskProvider(), raoData.getLoopflowCnecs());
-
-            // We may want to have a different interface for the first run and the successive runs if we do not wish to
-            // compute the PTDFs at every iteration.
         }
 
         return builder.build();
--- a/ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoUtil.java
+++ b/ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoUtil.java
@@ -87,7 +87,7 @@ public final class RaoUtil {
         }
     }
 
-    public static SystematicSensitivityInterface createSystematicSensitivityInterface(RaoParameters raoParameters, RaoData raoData, boolean withPtdfSensitivities) {
+    public static SystematicSensitivityInterface createSystematicSensitivityInterface(RaoParameters raoParameters, RaoData raoData, boolean withPtdfSensitivitiesForLoopFlows) {
 
         SystematicSensitivityInterface.SystematicSensitivityInterfaceBuilder builder = SystematicSensitivityInterface
             .builder()
@@ -95,7 +95,7 @@ public final class RaoUtil {
             .withFallbackParameters(raoParameters.getFallbackSensitivityAnalysisParameters())
             .withRangeActionSensitivities(raoData.getAvailableRangeActions(), raoData.getCnecs());
 
-        if (raoParameters.isRaoWithLoopFlowLimitation() && withPtdfSensitivities) {
+        if (raoParameters.isRaoWithLoopFlowLimitation() && withPtdfSensitivitiesForLoopFlows) {
             builder.withPtdfSensitivities(raoData.getGlskProvider(), raoData.getLoopflowCnecs());
         }
 
@@ -126,7 +126,7 @@ public final class RaoUtil {
     }
 
     private static MaxLoopFlowFiller createMaxLoopFlowFiller(RaoParameters raoParameters) {
-        return new MaxLoopFlowFiller(raoParameters.getLoopFlowConstraintAdjustmentCoefficient(), raoParameters.getLoopFlowViolationCost());
+        return new MaxLoopFlowFiller(raoParameters.getLoopFlowConstraintAdjustmentCoefficient(), raoParameters.getLoopFlowViolationCost(), raoParameters.getLoopFlowApproximationLevel());
     }
 
     private static IteratingLinearOptimizerParameters createIteratingParameters(RaoParameters raoParameters) {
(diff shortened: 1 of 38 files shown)


ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoUtil.java
```diff
@@ -87,20 +87,16 @@ public static void checkParameters(RaoParameters raoParameters, RaoInput raoInpu
         }
     }
 
-    public static SystematicSensitivityInterface createSystematicSensitivityInterface(RaoParameters raoParameters, RaoData raoData) {
+    public static SystematicSensitivityInterface createSystematicSensitivityInterface(RaoParameters raoParameters, RaoData raoData, boolean withPtdfSensitivities) {
```
baptsegu, line 90: raoParameters is not sufficient here ? Not sure there is a need for added a new argument

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/linear_optimisation/iterating_linear_optimizer/IteratingLinearOptimizer.java
```diff
@@ -83,6 +83,8 @@ public String optimize(RaoData raoData) {
         String optimizedVariantId;
         for (int iteration = 1; iteration <= parameters.getMaxIterations(); iteration++) {
             optimizedVariantId = cracVariantManager.cloneWorkingVariant();
+            raoData.getCracResultManager().copyCommercialFlowsBetweenVariants(cracVariantManager.getWorkingVariantId(), optimizedVariantId);
```
baptsegu, line 86: could we copy the results here if and only if the loopFlowApproximationLevel is FIXED_PTDF or UPDATE_PTDF_WITH_TOPO ?

baptsegu: And could we do this in the runSensitivityAndUpdateResults() function instead ? I think it would be more readable

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/Leaf.java
```diff
@@ -79,7 +79,8 @@ public String getMessage() {
         this.raoParameters = raoParameters;
         this.raoData = raoData;
         initialVariantId = raoData.getInitialVariantId();
-        systematicSensitivityInterface = RaoUtil.createSystematicSensitivityInterface(raoParameters, raoData);
+        systematicSensitivityInterface = RaoUtil.createSystematicSensitivityInterface(raoParameters, raoData,
```
baptsegu, line 82: Do not add a new argument here, raoParameter is sufficient

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/Leaf.java
```diff
@@ -157,7 +161,8 @@ void evaluate() {
                             systematicSensitivityInterface.run(raoData.getNetwork(), raoParameters.getObjectiveFunction().getUnit()));
 
                     if (raoParameters.isRaoWithLoopFlowLimitation()) {
-                        LoopFlowUtil.buildLoopFlowsWithLatestSensi(raoData, raoParameters.isLoopFlowApproximation());
+                        LoopFlowUtil.buildLoopFlowsWithLatestSensi(raoData,
```
baptsegu, line 164: same here, raoParameters are in raoData.getCracVariantManager

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/linear_optimisation/fillers/MaxLoopFlowFiller.java
```diff
@@ -96,42 +81,35 @@ private void buildLoopFlowConstraintsAndUpdateObjectiveFunction(RaoData raoData,
...
-                linearProblem.infinity(),
-                cnec
+                    0,
+                    linearProblem.infinity(),
+                    cnec
             );
 
-            // - MaxLoopFlow + LoopFlowShift <= flowVariable + loopflowViolationVariable <= POSITIVE_INF
+            double commercialFlow = cnec.getExtension(CnecResultExtension.class).getVariant(raoData.getWorkingVariantId()).getCommercialFlowInMW();
+            // - MaxLoopFlow + commercialFlow <= flowVariable + loopflowViolationVariable <= POSITIVE_INF
```
baptsegu, line 97: For the iterations >=2, the commercialFlow is set during the systematic analaysis of iteration n-1 : OK

For iteration 1, I'm not sure that the commercialFlow has been properly set. I think that we need an 'initial loop-flow computation' at the beginning of the iterationLinearRao, no ?

baptsegu (changes requested):
some small changes and one bigger question :rocket:

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/linear_optimisation/fillers/MaxLoopFlowFiller.java
pet-mit: At the beginning of the RAO, it is done in the InitialSensitivityAnalysis.
For each new Leaf, it is done in Leaf.evaluate() if we should update the PSTs with topo changes

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoUtil.java
pet-mit: Unfortunately it does not only depend on the contents of RaoParameters. This method is called in 3 places : 
- in the LinearOptimizer: then we should compute PTDFs if approximation level is ..._AND_PST
- in the Leaf, for the initial evaluation: then we should compute PTDFs if approximation level is UPDATE_PTDF_WITH_TOPO or UPDATE_PTDF_WITH_TOPO_AND_PST 
- in the Leaf, when creating the IteratingLinearOptimizer: then we should compute PTDFs if approx level is ..._AND_PST like in the linear optimizer case

baptsegu: ok I see :) 
There might be some duplicated code in the InitialSensitivityAnalysis that could be avoided now thanks to your new method signature. If you could have a look :angel:

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/linear_optimisation/fillers/MaxLoopFlowFiller.java
baptsegu: Indeed, perfect then !

baptsegu (approved):

PR #377, opened 2020-11-25
Make check of mnec/cnec optional

terriervik:

**Other information**:


terriervik:
@murgeyseb Am I allowed to do this? My input files contain no mention of cnec or mnec, and I would like to avoid this check in cleanCrac.

baptsegu:
Hello Viktor,

I see the idea of the PR, and technically am fine with it, but I am a bit annoyed with two things :
- using a rao-commons module outside of a RAO : which is actually a good sign because we are doing usefull things :monkey_face: but it might therefore deserved to be put in a non RAO related package : crac-utils ?
- changing the signature of the cleanCrac method, which for one parameter is fine, but which can become messy if we have to deal with many parameters for several specific uses.

I check with the rest of the team in tomorrow's daily.

PR #381, opened 2020-12-02
[fix github CI] add jaxb dependency

baptsegu:
Fix github CI

JAXB is apparently removed from java 11 and therefore need to be explicitly given in maven dependencies : https://stackoverflow.com/questions/52502189/java-11-package-javax-xml-bind-does-not-exist

I used the same version/dependencies than the ones used in powsybl-core java 11 profile :
https://github.com/powsybl/powsybl-core/blob/master/pom.xml


baptsegu:
already included

PR #383, opened 2020-12-04
update powsybl versions : core v3.8.0 rte-core v2.15.0

baptsegu:


distribution/src/main/resources/config.yml
```diff
@@ -1,78 +1,54 @@
-componentDefaultConfig:
-    SensitivityComputationFactory: com.rte_france.powsybl.hades2.sensitivity.Hades2SensitivityFactory
-    SensitivityFactorsProviderFactory: com.powsybl.sensitivity.JsonSensitivityFactorsProviderFactory
-    FlowDecompositionFactory: com.farao_community.farao.flow_decomposition.full_line_decomposition.FullLineDecompositionFactory
-    RaoComputationFactory: com.farao_community.farao.closed_optimisation_rao.ClosedOptimisationRaoFactory
-    FlowBasedComputationFactory: com.farao_community.farao.flowbased_computation.FlowBasedComputationFactoryImpl
-
 hades2:
-    homeDir: $HOME/farao-dep/runtime/hades2
+    homeDir: /home/farao/farao-dep/runtime/hades2
```
murgeyseb, line 2: why changing this, that will not work when using install.sh for anyone ?

murgeyseb (changes requested):

distribution/src/main/resources/config.yml
baptsegu: my bad, it's an unfortunate change of the config update. I correct it

murgeyseb (approved):

PR #385, opened 2020-12-08
filter onState usage rules whose state does not exist

baptsegu:

Add a new behaviour in the cleanCrac method : OnState usage rule based on a State which has been previously removed by the cleanCrac (typically because its contingency's network elements is not in the Network) are cleaned.


ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoInputHelper.java
```diff
@@ -83,44 +85,76 @@ public static void synchronize(Crac crac, Network network) {
...
+        // remove states and contingencies
+        removedContingencies.forEach(contingency -> crac.removeContingency(contingency.getId()));
+        removedStates.forEach(state -> crac.removeState(state.getId()));
 
         report.forEach(LOGGER::warn);
 
         return report;
     }
+
+    private static void checkUsageRules(RemedialAction<?> remedialAction, Set<State> removedStates, List<String> report) {
```
pet-mit, line 148: rename this to cleanUsageRules ?

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoInputHelper.java
```diff
@@ -83,44 +85,76 @@ public static void synchronize(Crac crac, Network network) {
...
+            report.add(String.format("[REMOVED] Remedial Action %s has no associated action. It is removed from the Crac", ra.getId()));
+            noValidAction.add(ra);
+        });
         noValidAction.forEach(networkAction -> crac.removeNetworkAction(networkAction.getId()));
 
-        // todo : remove State whose contingency does not exist anymore
-        // todo : remove onState remedial action whose state does not exist anymore
+        // remove On State usage rule with invalid state
+        crac.getRangeActions().forEach(ra -> checkUsageRules(ra, removedStates, report));
+        crac.getNetworkActions().forEach(ra -> checkUsageRules(ra, removedStates, report));
```
pet-mit, line 137: should we remove remedial actions with no usage rules left after this ? (if the RA only had one OnState usage rule that was removed by checkUsageRules)

pet-mit (changes requested):

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoInputHelper.java
baptsegu: This is something i was considering, but Farao actually handle well the case where a RemedialAction has no usage Rule : it will consider it UNAVAILABLE for all states.

I therefore prefer to keep these remedial actions in the Crac, then we still have it in the results and we can see that it was never activated. I allows to keep some more consistency between input/output without risking a crash of the RAO

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoInputHelper.java
baptsegu: you're right. I changed it.

ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/RaoInputHelper.java
pet-mit: okay !

pet-mit (approved):

PR #394, opened 2020-12-15
Handle reversed aliases

terriervik:

Bug fix

During the alias creation, the reversed cnecs were not handled correctly. Their thresholds were not reversed whereas the elementNames were added to the reversed network element.

This feature reversed the thresholds.

### DIFF
--- a/data/crac/crac-util/src/main/java/com/farao_community/farao/data/crac_util/CracAliasesCreator.java
+++ b/data/crac/crac-util/src/main/java/com/farao_community/farao/data/crac_util/CracAliasesCreator.java
@@ -8,6 +8,7 @@
 package com.farao_community.farao.data.crac_util;
 
 import com.farao_community.farao.data.crac_api.*;
+import com.farao_community.farao.data.crac_api.cnec.BranchCnec;
 import com.powsybl.iidm.network.Identifiable;
 import com.powsybl.iidm.network.*;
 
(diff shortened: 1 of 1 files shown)


data/crac/crac-util/src/main/java/com/farao_community/farao/data/crac_util/CracAliasesCreator.java
```diff
@@ -32,13 +33,42 @@ public void createAliases(Crac crac, Network network) {
...
+                    Optional<Double> max = branchThreshold.max();
+                    if (min.isPresent()) {
+                        branchThreshold.max() = Optional.of(-min.get());
+                    } else {
+                        branchThreshold.max() = Optional.of(Double.MAX_VALUE);
+                    }
+                    if (max.isPresent()) {
+                        branchThreshold.min() = Optional.of(-max.get());
+                    } else {
+                        branchThreshold.min() = Optional.of(-Double.MAX_VALUE);
```
terriervik, line 64: This doesn't work, what can I do?
(the pb is that branchThreshold.max() is a value, not a variable, so I can't write value = something else)

pet-mit:
Please provide a detailed description of the issue, with simple example data if possible

pet-mit:
Please provide a detailed description of the issue, with simple example data if possible

terriervik:
> Please provide a detailed description of the issue, with simple example data if possible

Do I get help faster if I do so?

data/crac/crac-util/src/main/java/com/farao_community/farao/data/crac_util/CracAliasesCreator.java
baptsegu: the max() method is a getter, so you cannot change the max() value.

Actually there is no setter for the min/max values of a Threshold. I don't really know if its intentional (it ensures that the threshold values cannot be changed, so it could be), or if it's just that we didn't have this need until now.

The easiest way to to this would be to add the setMin() and setMax() methods in the Threshold API and its implementations. Then you would be able to reverse the min and max values.

But i could be interesting to have also the opinion of TheMaskedTurtle who recently worked on the threshold --> maybe he intentionally decided to keep the min/max values finals.

baptsegu: We are facing very similar problems in CORE data-sets (inversions of from/to of the Cnecs + inversions of thresholds of the Cnecs). We decided to handle them "outside of farao", in a pre-processing script (which is not in an open-source repository).

Maybe you could directly use this script (it also handles some problems which exists with the transformers and tie-lines, that you might face in the future :s) -> I can help you use it.

Moreover, we are planning a refactoring of the Crac importers which will notably solve that kind of problems ! :) The idea would be that the importer takes a Crac **AND** a Network. And that all the mismatches between the Crac and the Network are dealt with during the import, and not in some wonky steps afterward.

Our first intention is to do it for CORE data, so with flowbased-document and UCTE network. --> so if you are using the same format, we will soon have something better for this...

By "soon", I mean 1-2 months :angel:

PR #395, opened 2020-12-15
Add curative RAO stop criterion

pet-mit:

Feature

All curative perimeters are optimized, even though in some cases the initial result is much better than the optimal result reached in the preventive perimeter.
From a minimum margin maximization perspective, optimizing curative perimeters is not necessary when the initial worst margin is better than the worst one already reached

In the new version, the stop criterion of the search tree is separated for preventive and curative.

1. "preventive-rao-stop-criterion" : can be "MIN_OBJECTIVE" (equivalent of old MAXIMUM_MARGIN) or "SECURE" (equivalent of old POSITIVE_MARGIN)
2. For curative RAO, the user can chose to optimize curative RAO performance by using one of three options in the "curative-rao-stop-criterion":
- MIN_OBJECTIVE: minimize the objective even if the optimal preventive objective was bested (same as current behavior)
- PREVENTIVE_OBJECTIVE: stop the curative RAO when the preventive objective is bested by "curative-rao-min-obj-improvement" (another new parameter, which you can set to zero if you don't care about a minimum improvement of the preventive objective)
- SECURE: stop the RAO when the situation is secure
- PREVENTIVE_OBJECTIVE_AND_SECURE: stop the curative RAO when the preventive objective is bested by "curative-rao-min-obj-improvement" AND the situation is secure

A new internal "TreeParameters" class is generated dynamically using these parameters and sent to the SearchTree. This class is not accessible to the user.

Update parameter files accordingly:
- Replace "stop-criterion" by "preventive-rao-stop-criterion" and "curative-rao-stop-criterion"
- Choose the two stop criteria depending on the behavior wanted


baptsegu (commented):
Functionally all is good to me.

However, I am starting to get a bit lost with the configuration.
As I understood it : 
* RaoParameters includes all parameters common to both RAO implementations, or which seem consistent to have in any RAO

* SearchTreeRaoParameters includes the parameters related to the RAO implementation SearchTreeRaoProvider : i.e. parameters related to an execution of this RAO which might includes several "Trees" (one for each perimeter)

* ??? : maybe we lack an internal parameter class which include the parameter for one "Tree" execution, is not accessible to the user (not in platformConfig, no I/O in JSON), but is managed by the SearchTreeRaoProvider when it runs its several "Trees"

ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/RaoParameters.java
```diff
@@ -32,6 +34,7 @@
  * @author Baptiste Seguinot {@literal <baptiste.seguinot at rte-france.com>}
  */
 public class RaoParameters extends AbstractExtendable<RaoParameters> {
+    static final Logger LOGGER = LoggerFactory.getLogger(RaoParameters.class);
```
baptsegu, line 37: not sure it is used (it is you only change in this file)

ra-optimisation/search-tree-rao/src/test/resources/parameters.json
```diff
@@ -29,7 +29,7 @@
   },
   "extensions" : {
     "SearchTreeRaoParameters" : {
-      "stop-criterion" : "MAXIMUM_MARGIN",
+      "preventive-stop-criterion" : "MAXIMUM_MARGIN",
```
baptsegu, line 32: MIN_OBJECTIVE

baptsegu (commented):
Thank you for the changes :) 

This new version looks fine to me, just some minor comments

ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/RaoParameters.java
pet-mit: true

ra-optimisation/search-tree-rao/src/test/resources/parameters.json
pet-mit: corrected, thanks

baptsegu (approved):

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/SearchTree.java
```diff
@@ -65,9 +59,10 @@ void initLeaves(RaoData raoData) {
         previousDepthOptimalLeaf = rootLeaf;
     }
 
-    public CompletableFuture<RaoResult> run(RaoData raoData, RaoParameters raoParameters) {
+    public CompletableFuture<RaoResult> run(RaoData raoData, RaoParameters raoParameters, TreeParameters treeParameters) {
```
alexandrem-rte, line 62: I would make TreeParameters an extension of RaoParameters. Thus you would not need to add this argument to this method.

pet-mit: we actually want to hide TreeParameters from the user. if we put it in RaoParameters, I think it might get confusing

PR #405, opened 2021-01-08
Add Z11 and relative margins to CNE export (#12.2)

pet-mit:

Feature

Z11 measurement (absolute PTDF sum for a given CNEC) is not exported in CNE file
The objective function (Z13 and Z15) is always equal to the opposite of the margin

Z11 is exported
The objective function is relative (divider by Z11) when the margin is positive

Whether the objective function is relative or not is stored in the CracResult extension of the CRAC. This requires us to correctly initialize CracResult variants, thus changing the constructor of RaoData
When creating a RaoData from scratch, users now have to set the boolean "relativePositiveMargins" (eventually using the objective function in RaoParameters, like it is done in the implemented RAOs)
When creating a RaoData using an existing variant, the constructor will automatically copy this boolean from the existing variant so the user doesn't have to set it.
This is why we now have two constructors for RaoData


baptsegu (commented):
You will have lots of conflicts with the master branch... but maybe not so complicated to solve. The RaoData now includes a RaoParameter, so it might be easier to handle the information on the objective function and to transfer it to the CracResultExtension.

Here you save the information of the relative margin (true/false). But later on I guess it could be interesting to store all the RaoParameters. It would be usefull to export the results, but also to explain them.

However we don't want the crac-result-extension package to depend from the rao-api package, it would be strange ...

The best way i think would be to : 
* put in the RaoResult the RaoParameters which was used by the Rao
* add the RaoResult in the input of the CNE exporter.

--> but this is pending for the refactoring on the Crac importer/exporters. We really need to do it quickly :)

pet-mit:
This implementation is too intrusive
An easier way will be possible after CRAC I/O refactoring
Meanwhile a simpler solution has been suggested : #422

PR #444, opened 2021-03-08
intra-day cse hybrid glsk of swissgrid - import and conversion

ghost:


phiedw (approved):

phiedw (changes requested):
Could you also do a small PR on farao-cucumber-tests to fix the compile errors please? (We use the GlskPointLinearGlskConverter.convert(...) method so we need to add the new integer attribute)

baptsegu:
similar features implemented in https://github.com/farao-community/farao-core/pull/587

PR #452, opened 2021-03-23
Flowbased performances enhancement

terriervik:
- [X] Docs have been added / updated (for bug fixes / features)

Performance enhancement

A sensi was computed for each curative state, even when no CRA is applied (= no computation was required, as it won't change the current results).

The curative states with no changes are filtered out.


terriervik:
I tested this on a real case, and I get exactly the same results, so I guess this PR is quite reliable! :champagne: :fireworks:

baptsegu (approved):

PR #464, opened 2021-05-18
Post-check RAO results

pet-mit:

Bug fixes

1. Aligned PSTs that have different initial setpoints are considered in the RAO. This makes the initial situation infeasible and can lead to violating virtual costs.
2. Given the separation between preventive and curative RAOs, the preventive RAO can degrade the margins on curative CNECs, and this margin may not be improved by the curative RAs. This may lead to a final solution worse than the initial one.
3. Virtual hubs are considered even if they are disconnected

1. Aligned PSTs with different initial setpoints are filtered out of the linear optimizer; thus they are not optimized and keep their initial setpoints.
2. At the end of the RAO, if the RAO has degraded a (curative) margin or virtual cost, the initial solution is kept as optimal.
3. Virtual hubs that are disconnected are not considered in the computations


baptsegu:
Implemented in another PR
https://github.com/farao-community/farao-core/pull/542

PR #473, opened 2021-06-04
Ignore PST filtering when TSO is not in "max-curative-pst-per-tso" map

pet-mit:

Bug fix

- PSTs belonging to TSOs absent from the "max-curative-pst-per-tso" map are filtered out
- In the search tree, PSTs used in previous depth are not taken into account into current depth's filtering

- When the TSO is absent from the map, it should be considered unconstrained. PSTs should not be filtered out.
- PSTs from previous depths are taken into account in case most limiting element changes between depths


baptsegu (commented):
code smells can be easily fix.

Otherwise all good for me :)

pet-mit:
> code smells can be easily fix.
> 
> Otherwise all good for me :)

true ! 
did you see that I added another fix ? (I forgot to add it to the PR description), there's a cucumber test that goes with it)

baptsegu (approved):

PR #478, opened 2021-06-10
LF computation : ignore LinearGlsk not connected to main network component (master)

pet-mit:
…ponent

Bug fix

When a LinearGlsk is not connected to the main network component, its PTDF is not null. Thus it is taken into account in the loopflow computations.

These LinearGlsk are now ignored.


loopflow-computation/src/main/java/com/farao_community/farao/loopflow_computation/LoopFlowComputationImpl.java
```diff
@@ -56,17 +63,33 @@ public LoopFlowResult calculateLoopFlows(Network network, SensitivityAnalysisPar
...
+                    return true;
+                }
+            } else {
+                Load load = network.getLoad(glsk);
+                if (load != null && load.getTerminal().getBusView().getBus().isInMainConnectedComponent()) {
+                    return true;
+                }
+            }
+        }
+        return false;
```
baptsegu, line 90: i would prefer to throw an exception in that case, as it is not suppose to happen and this case can "hide" some other problems in the previous steps.

baptsegu (changes requested):

loopflow-computation/src/main/java/com/farao_community/farao/loopflow_computation/LoopFlowComputationImpl.java
pet-mit: it would throw an exception even if all loads / gens exist, but are not connected to the main component ?

baptsegu: I was more thinking : 
if the load and the generator does not exist -> throw an exception (that case should be spotted in the GLSKs methods)
if one of them exist and is disconnected -> return false

pet-mit: alright ! i updated it with a random commit message :p

baptsegu (approved):

loopflow-computation/src/main/java/com/farao_community/farao/loopflow_computation/LoopFlowComputationImpl.java
baptsegu: :)

PR #481, opened 2021-06-14
Rollback RangeAction setpoints after degradation in iterating LP (v2.4 patch)

pet-mit:

bug fix

When an LP optimal solution degrades the cost (for example if it creates a virtual cost, only detected after sensi), the applied range actions after optim are not rolled back. This can lead to unexpected behavior (for example in controlling the max number of PSTs or CRAs)

Range actions are rolled back to their initial value


baptsegu (approved):
thanks for the bug fix :)

PR #544, opened 2021-11-10
replaced comparator lambda by stocking some values in a map

phiedw:

No

Perf increase

When computing most limiting elements, the comparator used by the stream.sorted() method has to recompute the margin for every comparison done.

We now store the margin for each flowCnec in a map, and use that map in the comparator.

**Other information**:


phiedw:
It might be even better to construct a TreeMap instead of a HashMap when computing the margins initially so the data is already sorted?
Even better would be doing some kind of insertion algorithm where we only store the k smallest values (should be O(kn) instead of O(nlog(n)) which is theoretically better since k values are always small where as n can be quite large)

baptsegu (approved):

PR #547, opened 2021-11-17
Attempt to render MIP results completely reproducible 

phiedw:

No

Bug fix

Running the same RAO twice in a row can give slightly different results

Running the same RAO should now give exactly the same results.
To do so we sort the sets of cnecs and range actions in the LP fillers, and round some values to a value representable with a double.

No

**Other information**:


baptsegu (approved):
:rocket:

PR #555, opened 2021-11-29
Fix MIP reproducibility

phiedw:

No

Bug fix

Very small changes in sensitivity/flow values can cause massive changes in the MIP resolution. To avoid this, values are rounded before being "inputed" to the MIP. The old patch didnt cover cases where we looked up variables, but we now correctly get our custom classes for all variable / constraints

No

**Other information**:

### DIFF
--- a/ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/linear_optimisation/FaraoMPSolver.java
+++ b/ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/linear_optimisation/FaraoMPSolver.java
@@ -18,7 +18,7 @@ import java.util.Map;
  */
 public class FaraoMPSolver extends MPSolver {
 
-    private static final double PRECISION = 1048576; //2^20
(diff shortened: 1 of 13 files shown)


ra-optimisation/rao-commons/src/main/java/com/farao_community/farao/rao_commons/linear_optimisation/FaraoMPSolver.java
```diff
@@ -7,37 +7,86 @@
...
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * @author Philippe Edwards {@literal <philippe.edwards at rte-international.com>}
  */
 public class FaraoMPSolver extends MPSolver {
 
-    private static final double PRECISION = 1048576; //2^20
+    private static final double PRECISION = 1073741824; //2^30
```
baptsegu, line 21: not easy to read. Could we have something like NUMBER_OF_TRUNCATED_BITS_FOR_DOUBLE_PRECISION (=30), or something more explicit ?

baptsegu (approved):

PR #559, opened 2021-12-15
json CRAC versioning + add TU to keep compatibility with old CRACs

baptsegu:
Add TU, to ensure that keep compatibility with previous versions of CRACs, when modifications are made in crac-io-json


data/crac-io/crac-io-json/src/main/java/com/farao_community/farao/data/crac_io_json/JsonSerializationConstants.java
```diff
@@ -22,6 +25,15 @@
 
     private JsonSerializationConstants() { }
 
+    public static final String CRAC_IO_VERSION = "1.1";
```
phiedw, line 28: maybe we could make it so we can set the version we want to export (and have the default set to the latest)

pet-mit (approved):

PR #571, opened 2022-01-11
Sensitivity provider parameter

pet-mit:

Feature

The sensitivity parameter is configured in the itools configuration. It it thus impossible for a single application to call two different sensitivity providers, depending on the case being treated.

A new "sensitivity-provider" string parameter in RaoParameters can be used to select the sensitivity provider to be used.


loopflow-computation/src/main/java/com/farao_community/farao/loopflow_computation/LoopFlowComputationImpl.java
```diff
@@ -41,11 +41,13 @@
     protected ReferenceProgram referenceProgram;
     protected Network network;
     protected Map<EICode, LinearGlsk> glskMap;
+    protected String sensitivityProvider;
 
-    public LoopFlowComputationImpl(ZonalData<LinearGlsk> glsk, ReferenceProgram referenceProgram, Network network) {
+    public LoopFlowComputationImpl(ZonalData<LinearGlsk> glsk, ReferenceProgram referenceProgram, Network network, String sensitivityProvider) {
```
baptsegu, line 46: i would put the new `sensitivityProider` argument in in the `calculateLoopFlows()` method instead, the method which also takes in argument a sensitivityAnalysisParameter.

Within the RAO, we only use the `buildLoopFlowsFromReferenceFlowAndPtdf()`, which rely on a previously calculated sensitivity result, and therefore which does not run internally a sensi

ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/parameters/RaoParameters.java
```diff
@@ -472,6 +492,7 @@ public static void load(RaoParameters parameters, PlatformConfig platformConfig)
                 parameters.setSolverSpecificParameters(config.getStringProperty("solver-specific-parameters", DEFAULT_SOLVER_SPECIFIC_PARAMETERS));
                 parameters.setPstOptimizationApproximation(config.getEnumProperty("pst-optimization-approximation", PstOptimizationApproximation.class, DEFAULT_PST_OPTIMIZATION_APPROXIMATION));
                 parameters.setForbidCostIncrease(config.getBooleanProperty("forbid-cost-increase", DEFAULT_FORBID_COST_INCREASE));
+                parameters.setSensitivityProvider(config.getStringProperty("sensitivity-provider", DEFAULT_SENSITIVITY_PROVIDER));
```
baptsegu, line 495: also add here the "load-flow-provider"

baptsegu (changes requested):

ra-optimisation/rao-api/src/main/java/com/farao_community/farao/rao_api/parameters/RaoParameters.java
pet-mit: fixed

loopflow-computation/src/main/java/com/farao_community/farao/loopflow_computation/LoopFlowComputationImpl.java
pet-mit: makes sense, i'll change this

baptsegu (approved):
thanks for the changes !

PR #644, opened 2022-08-09
Cnec on dangling line

pet-mit:

Feature

The sensitivity providers cannot compute flows and sensitivities on dangling lines. Thus CNECs defined on dangling lines cause the RAO to crash.

Now, these CNECs are filtered out of the sensitivity computation and their flows and sensitivities are implicitly considered equal to zero. This is okay because when only a dangling line exists (instead of being part of a tie line), it should be considered as a disconnected tie line.


pet-mit:
Now handled by PowSyBl

PR #645, opened 2022-08-09
Cim border

pet-mit:

Feature

The CIM importer imports all TimeSeries in CRAC file

User can now specify which TimeSeries to import, by putting their mrIDs in the new CimCreationParameters field


phiedw (commented):
One risk I see is a user creating a CimCracCreationParameters with an empty set for the timeSeriesMrids (or even a null value, this seems to throw an error when deserializing) instead of not putting the field at all. Also, if for some reason at some point we want to be able to overwrite a json with another json, and the first json had a value for timeSeriesMrids, then we cannot over write that value at all.

I think that to solve that, we should consider an empty set as meaning: "use all the time series in the crac". This would mean either replacing all the null checks with empty checks and initializing the set as an empty set in the parameters, or just adding empty checks to the current null checks.

phiedw (approved):

PR #653, opened 2022-09-01
Fix PRAO2 run condition

pet-mit:

Bug fix

The 2nd preventive RAO run condition checks if any curative cost is worse than the initial cost.
The fall-back condition checks if max(functional cost) + sum(virtual costs) is worse than the initial cost.
These conditions are not coherent and the second preventive RAO is not always run when it should be.

Both conditions now check max + sum


ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/castor/algorithm/CastorFullOptimization.java
```diff
@@ -379,7 +379,7 @@ static boolean shouldRunSecondPreventiveRao(RaoParameters raoParameters, Objecti
      * Returns true if any result has a positive functional cost
      */
     private static boolean isAnyResultUnsecure(Collection<OptimizationResult> results) {
-        return results.stream().anyMatch(optimizationResult -> optimizationResult.getFunctionalCost() >= 0);
+        return results.stream().anyMatch(optimizationResult -> optimizationResult.getFunctionalCost() >= 0 || optimizationResult.getVirtualCost() > 1e-6);
```
phiedw, line 382: It would be more consistent to put "> 1e-6" on both conditions (functional cost and virtual cost)

phiedw (changes requested):
I believe the goal of the curative stop criterion PREVENTIVE_OBJECTIVE is to reduce the global cost of the RAO (ie the worst functional cost over all perimeters + the sum of all virtual costs). However in that case, we call isAnyCurativeWorseThanPreventive to check if we should run a 2nd preventive. This method checks costs perimeter per perimeter, so if a perimeter has a large min margin but a positive virtual cost, it may not consider a second preventive is required. But this virtual cost will be taken into account when computing the overall cost, and it may have been solved by a second preventive.
Should we change the isAnyCurativeWorseThanPreventive to check virtual costs too? (maybe it would then need a new name too)

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/castor/algorithm/CastorFullOptimization.java
pet-mit: I wanted to match the SearchTree stop criterion (0 is not considered secure)

pet-mit:
> I believe the goal of the curative stop criterion PREVENTIVE_OBJECTIVE is to reduce the global cost of the RAO (ie the worst functional cost over all perimeters + the sum of all virtual costs). However in that case, we call isAnyCurativeWorseThanPreventive to check if we should run a 2nd preventive. This method checks costs perimeter per perimeter, so if a perimeter has a large min margin but a positive virtual cost, it may not consider a second preventive is required. But this virtual cost will be taken into account when computing the overall cost, and it may have been solved by a second preventive. Should we change the isAnyCurativeWorseThanPreventive to check virtual costs too? (maybe it would then need a new name too)

Tricky question. I think there are two things here.
Maybe we should, instead, change the "COST_INCREASE" 2P run condition to something like "COST_INCREASE_OR_VIRTUAL_COST". Then we'll check virtual costs no matter the curative stop condition (we keep those checks as is)

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/castor/algorithm/CastorFullOptimization.java
phiedw: Okay. I forgot virtual costs cannot go below 0 whereas functional costs can, so it indeed makes sense to have different conditions

phiedw (approved):

PR #752, opened 2023-04-17
CoreCne : adapt to only use FlowResult in MW

bqth29:
## Context

In order to limit the size of the `RaoResult` files for the CORE region, only the flow and margin results in megawatts should appear (not those in amperes). On the other hand, the Cne exporter must be able to retrieve the thresholds in amperes from these megawatt results only.

### Tasks

- [x] Allow the reading of megawatts only
- [x] Update the RaoResults JSON files in the cucumber tests by removing flow measurements in megawatts (/amperes for SWE)

### Comments

The `getCnecFlow` method of the `CoreCneCnecsCreator` class was refactored such that it no longer takes the `Unit unit` argument. In exchange, it systematically calls the `RaoResult.getFlow` method with the `Unit unit` argument being `Unit.MEGAWATT`.


pet-mit (approved):
Approved
To test & merge after https://github.com/farao-community/farao-core/pull/754

PR #755, opened 2023-04-24
Monitor network copies in search tree

bqth29:

The PR provides a better management of the number of copies of the network (`MultipleNetworkPool` class) by limiting the number of copies by the number of network actions actually available.

The parallelised RAO is now supposed to be faster since no useless copies of the network are created.

The class now has a `addNetworkClones(int numberOfClones)` method that creates copies of the network. At present, there is no copy of the network initially in the Blocking Queue so a test is performed before calling a `NetworkPool` to ensure that the queue is not empty (which would result in an infinite loop).

The method `SearchTree.updateOptimalLeafWithNextDepthBestLeaf` was refactored since the `network` argument is no longer used because the `AbstractNetworkPool` know has a `network` attribute.

**Other information**:

### DIFF
--- a/util/src/main/java/com/farao_community/farao/util/AbstractNetworkPool.java
+++ b/util/src/main/java/com/farao_community/farao/util/AbstractNetworkPool.java
@@ -29,11 +29,11 @@ public abstract class AbstractNetworkPool extends ForkJoinPool implements AutoCl
     protected String networkInitialVariantId;
     protected Set<String> baseNetworkVariantIds;
 
-    public static AbstractNetworkPool create(Network network, String targetVariant, int parallelism) {
+    public static AbstractNetworkPool create(Network network, String targetVariant, int parallelism, boolean initClones) {
         if (parallelism == 1) {
             return new SingleNetworkPool(network, targetVariant);
         } else {
-            return new MultipleNetworkPool(network, targetVariant, parallelism);
+            return new MultipleNetworkPool(network, targetVariant, parallelism, initClones);
         }
     }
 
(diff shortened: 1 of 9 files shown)


util/src/main/java/com/farao_community/farao/util/AbstractNetworkPool.java
```diff
@@ -24,6 +25,9 @@ public abstract class AbstractNetworkPool extends ForkJoinPool implements AutoCl
     // State used to save initial content of target variant.
     // Useful when targetVariant equals VariantManagerConstants.INITIAL_VARIANT_ID
     protected final String stateSaveVariant;
+    protected Network network;
+    protected String networkInitialVariantId;
+    protected Set<String> baseNetworkVariantIds;
 
     public static AbstractNetworkPool create(Network network, String targetVariant, int parallelism) {
```
pet-mit, line 32: how about adding a 4th boolean argument "initClones". if set to true, the clones would be added in constructor (used in all classes except search tree). if set to false, then the user of the class should call addClones explicitely. This will at least make it lighter for classes other than search tree

util/src/main/java/com/farao_community/farao/util/MultipleNetworkPool.java
```diff
@@ -21,16 +20,49 @@
...
+    @Override
+    public int getNetworkNumberOfClones() {
+        // The number of clones includes the original network itself
+        return networkNumberOfClones;
     }
 
     @Override
-    protected void initAvailableNetworks(Network network) {
-        TECHNICAL_LOGS.debug("Filling network pool with copies of network '{}' on variant '{}'", network.getId(), targetVariant);
+    public void addNetworkClones(int numberOfClones) {
```
pet-mit, line 47: rename this to initClones(int desiredNumberOfClones)
inside method, add addedClones = min(parallelism, desiredNumberOfClones) - networkNumberOfClones

pet-mit (changes requested):

pet-mit (approved):

PR #756, opened 2023-04-24
Create less ORTools interfaces

pet-mit:

Feature

The iterating linear optimizer creates a new MPSolver every time it builds a LinearProblem. This is not necessary since it can add an overhead + it generates a banner log on each creation.

Now interfaces are created once and re-used if available
(PS : unfortunately CBC interface does not seem to handle well the call to reset(), which allows resetting varaibles & constraints before re-using the MPSolver. It has been excluded from this dev)


pet-mit:
impacts performance

PR #758, opened 2023-05-04
add limit to pst range for sensi changes

MartinBelthle:

Feature

The Rao currently assumes that the impact of a PST on a CNEC is linear. But in AC mode, this approximation is not true.
Therefore, the MIP can choose a solution which is in fact worse than the previous iteration. Currently, when this happens the MIP stops and keeps the best iteration (the last one).  

**What is the new behavior?**

This PR allows the MIP to continue even though it found a worse solution. It keeps in mind the best solution and continue until it reaches max_iteration. 

Moreover, we diminish the pst relative variation at each iteration to avoid the MIP to behave as seen on the left picture below. 

parabole

The factor 2/3 has been chosen to allow, if needed, the PST to come back to its initial tap. Here is an example of the new behavior:
The PST has 32 taps (-16 to +16). In initial state, its tap position is -16. After the 1st iteration, its tap position is +16 and the MIP has worsen the result. It will run 2nd iteration but for this iteration the PST can only go from tap -5.3 (tap 16 - 32 taps * 2/3) to tap +16. And for the next iterations, **its admissible range will shrink to : previous iteration tap +- 32 taps * (2/3)^current iteration number.**

### DIFF
--- a/ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/commons/parameters/TreeParameters.java
+++ b/ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/commons/parameters/TreeParameters.java
@@ -30,18 +30,18 @@ public final class TreeParameters {
     private final double targetObjectiveValue;
     private final int maximumSearchDepth;
     private final int leavesInParallel;
-    private final boolean capPstVariation;
+    private final boolean decreasePstRange;
 
     public TreeParameters(StopCriterion stopCriterion,
                            double targetObjectiveValue,
                            int maximumSearchDepth,
                            int leavesInParallel,
-                          boolean capPstVariation) {
+                          boolean decreasePstRange) {
         this.stopCriterion = stopCriterion;
         this.targetObjectiveValue = targetObjectiveValue;
         this.maximumSearchDepth = maximumSearchDepth;
         this.leavesInParallel = leavesInParallel;
-        this.capPstVariation = capPstVariation;
+        this.decreasePstRange = decreasePstRange;
     }
 
     public StopCriterion getStopCriterion() {
@@ -60,23 +60,30 @@ public final class TreeParameters {
         return leavesInParallel;
     }
 
-    public boolean getCapPstVariation() {
-        return capPstVariation;
+    public boolean getDecreasePstRange() {
+        return decreasePstRange;
     }
 
     public static TreeParameters buildForPreventivePerimeter(RaoParameters parameters) {
-        StopCriterion stopCriterion = StopCriterion.MIN_OBJECTIVE;
-        if (parameters.getObjectiveFunctionParameters().getPreventiveStopCriterion().equals(ObjectiveFunctionParameters.PreventiveStopCriterion.SECURE)) {
-            stopCriterion = StopCriterion.AT_TARGET_OBJECTIVE_VALUE;
-        } else if (!parameters.getObjectiveFunctionParameters().getPreventiveStopCriterion().equals(ObjectiveFunctionParameters.PreventiveStopCriterion.MIN_OBJECTIVE)) {
-            throw new FaraoException("Unknown preventive stop criterion: " + parameters.getObjectiveFunctionParameters().getPreventiveStopCriterion());
+        RangeActionsOptimizationParameters.PstRangeDecrease pstRangeDecrease = parameters.getRangeActionsOptimizationParameters().getPstRangeDecrease();
+        boolean decreasePstRange = pstRangeDecrease.equals(RangeActionsOptimizationParameters.PstRangeDecrease.ENABLED_IN_FIRST_PRAO_AND_CRAO) ||
+            pstRangeDecrease.equals(RangeActionsOptimizationParameters.PstRangeDecrease.ENABLED);
+        switch (parameters.getObjectiveFunctionParameters().getPreventiveStopCriterion()) {
+            case MIN_OBJECTIVE:
+                return new TreeParameters(StopCriterion.MIN_OBJECTIVE,
+                    0.0, // value does not matter
+                    parameters.getTopoOptimizationParameters().getMaxSearchTreeDepth(),
+                    parameters.getMultithreadingParameters().getPreventiveLeavesInParallel(),
+                    decreasePstRange);
+            case SECURE:
+                return new TreeParameters(StopCriterion.AT_TARGET_OBJECTIVE_VALUE,
+                    0.0, // secure
+                    parameters.getTopoOptimizationParameters().getMaxSearchTreeDepth(),
+                    parameters.getMultithreadingParameters().getPreventiveLeavesInParallel(),
+                    decreasePstRange);
+            default:
+                throw new FaraoException("Unknown preventive stop criterion: " + parameters.getObjectiveFunctionParameters().getPreventiveStopCriterion());
         }
-        boolean capPstVariation = !parameters.getRangeActionsOptimizationParameters().getPstVariationGradualDecrease().equals(RangeActionsOptimizationParameters.PstRangeDecrease.DISABLED);
-        return new TreeParameters(stopCriterion,
-            0.0,
-            parameters.getTopoOptimizationParameters().getMaxSearchTreeDepth(),
-            parameters.getMultithreadingParameters().getPreventiveLeavesInParallel(),
-            capPstVariation);
     }
 
     public static TreeParameters buildForCurativePerimeter(RaoParameters parameters, Double preventiveOptimizedCost) {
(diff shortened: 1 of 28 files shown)


ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/linear_optimisation/algorithms/linear_problem/LinearProblem.java
```diff
@@ -128,8 +128,12 @@ public FaraoMPVariable getRangeActionSetpointVariable(RangeAction<?> rangeAction
         return solver.getVariable(rangeActionSetpointVariableId(rangeAction, state));
     }
 
-    public FaraoMPConstraint addRangeActionRelativeSetpointConstraint(double lb, double ub, RangeAction<?> rangeAction, State state) {
-        return solver.makeConstraint(lb, ub, rangeActionRelativeSetpointConstraintId(rangeAction, state));
+    public FaraoMPConstraint addRangeActionRelativeSetpointConstraint(double lb, double ub, RangeAction<?> rangeAction, State state, String string) {
```
pet-mit, line 131: either add a whole new constraint, or use a boolean (better : enum) instead of string. using string is too liberating

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/linear_optimisation/algorithms/linear_problem/LinearProblemIdGenerator.java
```diff
@@ -60,8 +60,8 @@ public static String rangeActionSetpointVariableId(RangeAction<?> rangeAction, S
         return rangeAction.getId() + SEPARATOR + state.getId() + SEPARATOR + SET_POINT + SEPARATOR + VARIABLE_SUFFIX;
     }
 
-    public static String rangeActionRelativeSetpointConstraintId(RangeAction<?> rangeAction, State state) {
-        return rangeAction.getId() + SEPARATOR + state.getId() + SEPARATOR + RELATIVE + SEPARATOR + SET_POINT + SEPARATOR + CONSTRAINT_SUFFIX;
+    public static String rangeActionRelativeSetpointConstraintId(RangeAction<?> rangeAction, State state, String string) {
```
pet-mit, line 63: use enum or boolean

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/linear_optimisation/parameters/IteratingLinearOptimizerParameters.java
```diff
@@ -32,6 +32,7 @@ public final class IteratingLinearOptimizerParameters {
     private final RangeActionsOptimizationParameters.LinearOptimizationSolver solverParameters;
 
     private final int maxNumberOfIterations;
+    private final boolean capPstVariation;
```
pet-mit, line 35: rename this attribute

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/linear_optimisation/algorithms/fillers/CoreProblemFiller.java
```diff
@@ -240,6 +243,38 @@ private void buildRangeActionConstraints(LinearProblem linearProblem) {
...
+        double previousSetPointValue = rangeActionActivationResult.getOptimizedSetpoint(rangeAction, state);
+        List<Double> minAndMaxAbsoluteAndRelativeSetpoints = getMinAndMaxAbsoluteAndRelativeSetpoints(rangeAction);
+        double minAbsoluteSetpoint = Math.max(minAndMaxAbsoluteAndRelativeSetpoints.get(0), -LinearProblem.infinity());
+        double maxAbsoluteSetpoint = Math.min(minAndMaxAbsoluteAndRelativeSetpoints.get(1), LinearProblem.infinity());
+        double setPointRange = Math.abs(minAbsoluteSetpoint) + Math.abs(maxAbsoluteSetpoint);
+        double constrainedSetPointRange = setPointRange * Math.pow(RANGE_DIMINUTION_RATE, iteration);
+        FaraoMPConstraint iterativeShrink = linearProblem.getRangeActionRelativeSetpointConstraint(rangeAction, state, "iterative_shrink");
+        if (iterativeShrink != null) {
+            iterativeShrink.setLb(previousSetPointValue - constrainedSetPointRange);
+            iterativeShrink.setUb(previousSetPointValue + constrainedSetPointRange);
```
pet-mit, line 266: shouldn't this have a 0.5 coefficient? 
```suggestion
            iterativeShrink.setLb(previousSetPointValue - 0.5 * constrainedSetPointRange);
            iterativeShrink.setUb(previousSetPointValue + 0.5 * constrainedSetPointRange);
```

pet-mit (changes requested):

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/linear_optimisation/algorithms/fillers/CoreProblemFiller.java
```diff
@@ -240,6 +243,38 @@ private void buildRangeActionConstraints(LinearProblem linearProblem) {
         );
     }
 
+    private void updateRangeActionConstraints(LinearProblem linearProblem, RangeActionActivationResult rangeActionActivationResult) {
+        iteration++;
+        optimizationContext.getRangeActionsPerState().entrySet().stream()
+            .sorted(Comparator.comparingInt(e -> e.getKey().getInstant().getOrder()))
```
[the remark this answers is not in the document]
MartinBelthle: I did the same thing as in the method `buildRangeActionConstraints` but in fact i do not understand know why it was done. If there is no reason, i can remove it from both methods.

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/linear_optimisation/algorithms/fillers/CoreProblemFiller.java
MartinBelthle: I do not think we want to introduce a 0.5 coefficient because we chose the factor 2/3 on purpose to allow the PST to come back to its initial tap. If we add this coefficient, it will not be possible anymore.

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/commons/parameters/TreeParameters.java
```diff
@@ -56,21 +60,23 @@ public int getLeavesInParallel() {
         return leavesInParallel;
     }
 
+    public boolean getCapPstVariation() {
+        return capPstVariation;
+    }
```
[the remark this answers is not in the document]
pet-mit: we usually avoid exposing methods that are not needed in the main part of the code. this makes the design of the classes cleaner.
If you need to test the value, consider another way, or at least making the getter private package. If it's not possible, then you can keep the getter, it's not a big issue.

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/linear_optimisation/algorithms/fillers/CoreProblemFiller.java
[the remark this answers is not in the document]
pet-mit: when building new constraints, we try to always build them in a fixed order. building the optimization matrix in the same order always is important for the solver to produce the same solution on every run (when the optimization matrix is sorted differently, the solver can take a different path in B&B).
This is why artifacts are sorted before running buildRangeActionConstraints.
However, when changing coefficients on an existing constraint (which you are doing in the update method), you do not need to follow a specific order.

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/linear_optimisation/algorithms/fillers/CoreProblemFiller.java
pet-mit: OK, I hadn't understood this design choice

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/commons/parameters/TreeParameters.java
```diff
@@ -30,18 +30,18 @@ public enum StopCriterion {
...
     private final int leavesInParallel;
-    private final boolean capPstVariation;
+    private final boolean decreasePstRange;
 
     public TreeParameters(StopCriterion stopCriterion,
                            double targetObjectiveValue,
                            int maximumSearchDepth,
                            int leavesInParallel,
-                          boolean capPstVariation) {
+                          boolean decreasePstRange) {
```
pet-mit, line 39:
```suggestion
                          boolean shrinkPstRange) {
```

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/linear_optimisation/algorithms/IteratingLinearOptimizer.java
```diff
@@ -94,13 +94,13 @@ public static LinearOptimizationResult optimize(IteratingLinearOptimizerInput in
                     iteration,
                     input.getObjectiveFunction()
             );
-            lastResult = currentResult;
+            previousResult = currentResult;
 
-            Map<Boolean, IteratingLinearOptimizationResultImpl> mipShouldStop = updateMipAndLogResult(parameters.getCapPstVariation(), linearProblem, input, iteration, currentResult, bestResult);
-            if (mipShouldStop.containsKey(Boolean.TRUE)) {
+            Pair<IteratingLinearOptimizationResultImpl, Boolean> mipShouldStop = updateBestResultAndCheckStopCondition(parameters.getDecreasePstRange(), linearProblem, input, iteration, currentResult, bestResult);
+            if (mipShouldStop.getValue().equals(Boolean.TRUE)) {
```
pet-mit, line 100: The concept of key & value has no sense here. use left & right instead
Also you don't need the ".equals(Boolean.TRUE)"
```suggestion
            if (mipShouldStop.getRight()) {
```

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/linear_optimisation/algorithms/IteratingLinearOptimizer.java
```diff
@@ -94,13 +94,13 @@ public static LinearOptimizationResult optimize(IteratingLinearOptimizerInput in
...
                 return bestResult;
             } else {
-                bestResult = mipShouldStop.get(Boolean.FALSE);
+                bestResult = mipShouldStop.getKey();
```
pet-mit, line 103:
```suggestion
                bestResult = mipShouldStop.getLeft();
```

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/linear_optimisation/algorithms/linear_problem/LinearProblem.java
```diff
@@ -58,6 +58,22 @@ public enum BoundExtension {
         UPPER_BOUND
     }
 
+    public enum PstRangeDecrease {
```
pet-mit, line 61:
```suggestion
    public enum PstRangeShrinking {
```

pet-mit (changes requested):

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/commons/parameters/TreeParameters.java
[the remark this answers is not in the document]
MartinBelthle: I use this method in `Leaf.java` which doesn't belong to the same package so i left it public (as `getStopCriterion`, `getLeavesInParallel` etc. that are used in `SearchTree.java`)

pet-mit (approved):

PR #766, opened 2023-06-08
Better handling of max RA

bqth29:

**Does this PR already have an issue describing the problem ?**

[FAR-832]

This PR introduces 2 things :
-  It takes over the behavior that existed prior to the [#730](https://github.com/farao-community/farao-core/pull/730) PR:
_In search-tree, when range actions are applied at a given leaf, they are kept at their optimal set-points in children leaves. This improves search-tree computation performance._
- When `max_CRA` parameters are defined, network actions can now be selected by the search Tree.

**What is the current behaviour?**

1) In the search tree, at the beginning of every leaf, range actions are reset to their pre-perimeter set-points.
2) If any `max_CRA` constraint is reached at a given search Tree depth, the research stops.

**What is the new behavior ?**

1) As explained above, when range actions are applied at a given leaf, they are kept at their optimal set-points in children leaves
Therefore, the search tree is now faster and due to its quicker convergence, it can produce best result when the max number of MIP iterations is low.
2)  If `max_CRA` is reached at any depth of the tree, the search Tree can now reset range Actions set-points to their pre-perimeter value to allow the evaluation of network actions while respecting the `max_CRA` constraints.
It allows the RAO to find better solutions.

### DIFF
--- a/ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/Leaf.java
+++ b/ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/Leaf.java
@@ -70,7 +70,7 @@ public class Leaf implements OptimizationResult {
     private final Set<NetworkAction> appliedNetworkActionsInPrimaryState;
     private final AppliedRemedialActions appliedRemedialActionsInSecondaryStates; // for 2nd prev
     private Network network;
-    private final RangeActionActivationResult prePerimeterActivationResult;
+    private final RangeActionActivationResult raActivationResultFromParentLeaf;
     private final RangeActionSetpointResult prePerimeterSetpoints;
 
     /**
@@ -92,12 +92,12 @@ public class Leaf implements OptimizationResult {
          Network network,
          Set<NetworkAction> alreadyAppliedNetworkActionsInPrimaryState,
          NetworkActionCombination newCombinationToApply,
-         RangeActionActivationResult prePerimeterActivationResult,
+         RangeActionActivationResult raActivationResultFromParentLeaf,
          RangeActionSetpointResult prePerimeterSetpoints,
          AppliedRemedialActions appliedRemedialActionsInSecondaryStates) {
         this.optimizationPerimeter = optimizationPerimeter;
         this.network = network;
-        this.prePerimeterActivationResult = prePerimeterActivationResult;
+        this.raActivationResultFromParentLeaf = raActivationResultFromParentLeaf;
         this.prePerimeterSetpoints = prePerimeterSetpoints;
         if (!Objects.isNull(newCombinationToApply)) {
             this.appliedNetworkActionsInPrimaryState = Stream.concat(
@@ -152,7 +152,7 @@ public class Leaf implements OptimizationResult {
     void evaluate(ObjectiveFunction objectiveFunction, SensitivityComputer sensitivityComputer) {
         if (status.equals(Status.EVALUATED)) {
             TECHNICAL_LOGS.debug("Leaf has already been evaluated");
-            preOptimObjectiveFunctionResult = objectiveFunction.evaluate(preOptimFlowResult, prePerimeterActivationResult, preOptimSensitivityResult, preOptimSensitivityResult.getSensitivityStatus());
+            preOptimObjectiveFunctionResult = objectiveFunction.evaluate(preOptimFlowResult, raActivationResultFromParentLeaf, preOptimSensitivityResult, preOptimSensitivityResult.getSensitivityStatus());
             return;
         }
         TECHNICAL_LOGS.debug("Evaluating {}", this);
@@ -164,7 +164,7 @@ public class Leaf implements OptimizationResult {
         }
         preOptimSensitivityResult = sensitivityComputer.getSensitivityResult();
         preOptimFlowResult = sensitivityComputer.getBranchResult(network);
-        preOptimObjectiveFunctionResult = objectiveFunction.evaluate(preOptimFlowResult, prePerimeterActivationResult, preOptimSensitivityResult, preOptimSensitivityResult.getSensitivityStatus());
+        preOptimObjectiveFunctionResult = objectiveFunction.evaluate(preOptimFlowResult, raActivationResultFromParentLeaf, preOptimSensitivityResult, preOptimSensitivityResult.getSensitivityStatus());
         status = Status.EVALUATED;
     }
 
@@ -200,7 +200,7 @@ public class Leaf implements OptimizationResult {
                     .withPreOptimizationFlowResult(preOptimFlowResult)
                     .withPreOptimizationSensitivityResult(preOptimSensitivityResult)
                     .withPreOptimizationAppliedRemedialActions(appliedRemedialActionsInSecondaryStates)
-                    .withRaActivationFromParentLeaf(prePerimeterActivationResult)
+                    .withRaActivationFromParentLeaf(raActivationResultFromParentLeaf)
                     .withObjectiveFunction(searchTreeInput.getObjectiveFunction())
                     .withToolProvider(searchTreeInput.getToolProvider())
                     .build();
@@ -231,7 +231,7 @@ public class Leaf implements OptimizationResult {
 
     private void resetPreOptimRangeActionsSetpoints(OptimizationPerimeter optimizationContext) {
         optimizationContext.getRangeActionsPerState().forEach((state, rangeActions) ->
-                rangeActions.forEach(ra -> ra.apply(network, prePerimeterActivationResult.getOptimizedSetpoint(ra, state))));
+                rangeActions.forEach(ra -> ra.apply(network, raActivationResultFromParentLeaf.getOptimizedSetpoint(ra, state))));
     }
 
     private RangeActionLimitationParameters getRaLimitationParameters(OptimizationPerimeter context, SearchTreeParameters parameters) {
@@ -287,7 +287,7 @@ public class Leaf implements OptimizationResult {
 
     public RangeActionActivationResult getRangeActionActivationResult() {
         if (status == Status.EVALUATED) {
-            return prePerimeterActivationResult;
+            return raActivationResultFromParentLeaf;
         } else if (status == Status.OPTIMIZED) {
             return postOptimResult.getRangeActionActivationResult();
         } else {
@@ -323,7 +323,7 @@ public class Leaf implements OptimizationResult {
     long getNumberOfActivatedRangeActions() {
         if (status == Status.EVALUATED) {
             return (long) optimizationPerimeter.getRangeActionsPerState().keySet().stream()
-                    .mapToDouble(s -> prePerimeterActivationResult.getActivatedRangeActions(s).size())
+                    .mapToDouble(s -> raActivationResultFromParentLeaf.getActivatedRangeActions(s).size())
                     .sum();
         } else if (status == Status.OPTIMIZED) {
             return (long) optimizationPerimeter.getRangeActionsPerState().keySet().stream()
@@ -480,7 +480,7 @@ public class Leaf implements OptimizationResult {
     @Override
     public Set<RangeAction<?>> getRangeActions() {
         if (status == Status.EVALUATED) {
-            return prePerimeterActivationResult.getRangeActions();
+            return raActivationResultFromParentLeaf.getRangeActions();
         } else if (status == Status.OPTIMIZED) {
             return postOptimResult.getRangeActions();
         } else {
@@ -491,7 +491,7 @@ public class Leaf implements OptimizationResult {
     @Override
     public Set<RangeAction<?>> getActivatedRangeActions(State state) {
         if (status == Status.EVALUATED) {
-            return prePerimeterActivationResult.getActivatedRangeActions(state);
+            return raActivationResultFromParentLeaf.getActivatedRangeActions(state);
         } else if (status == Status.OPTIMIZED) {
             return postOptimResult.getActivatedRangeActions(state);
         } else {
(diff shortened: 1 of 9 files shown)


ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTreeBloomer.java
```diff
@@ -146,13 +162,32 @@ List<NetworkActionCombination> removeCombinationsWhichExceedMaxNumberOfRa(List<N
         return filteredNaCombinations;
     }
 
-    List<NetworkActionCombination> removeCombinationsWhichExceedMaxNumberOfRaPerTso(List<NetworkActionCombination> naCombinations, Leaf fromLeaf) {
+    Map<NetworkActionCombination, Boolean> removeCombinationsWhichExceedMaxNumberOfRaPerTso(Map<NetworkActionCombination, Boolean> naCombinations, Leaf fromLeaf) {
```
Godelaine, line 165: could you try to refactor to do this more efficiently ?

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTreeBloomer.java
```diff
@@ -146,13 +162,32 @@ List<NetworkActionCombination> removeCombinationsWhichExceedMaxNumberOfRa(List<N
...
 
+        Map<NetworkActionCombination, Boolean> filteredNaCombinations = new HashMap<>();
```
Godelaine, line 167: update doc

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/Leaf.java
```diff
@@ -92,11 +92,12 @@ public enum Status {
          Network network,
          Set<NetworkAction> alreadyAppliedNetworkActionsInPrimaryState,
          NetworkActionCombination newCombinationToApply,
+         RangeActionActivationResult prePerimeterActivationResult,
```
Godelaine, line 95: add a comment explaining why sometimes prePerimeterActivationResult will be different than new RangeActionActivationResultimpl(preperimeterSetpoints)

ra-optimisation/search-tree-rao/src/test/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTreeBloomerTest.java
```diff
@@ -217,128 +225,164 @@ void testRemoveCombinationsWhichExceedMaxNumberOfRaPerTso() {
...
     }
 
     @Test
     void testRemoveNetworkActionsFarFromMostLimitingElement() {
 
         // arrange naCombination list
-        List<NetworkActionCombination> naCombinations = List.of(indFr2, indDe1, indBe1, indNl1, indNlBe, indFrDe, indDeNl, comb3Be, comb2De, comb2FrDeBe, comb2BeNl);
+        List<NetworkActionCombination> listOfNaCombinations = List.of(indFr2, indDe1, indBe1, indNl1, indNlBe, indFrDe, indDeNl, comb3Be, comb2De, comb2FrDeBe, comb2BeNl);
+        Map<NetworkActionCombination, Boolean> naCombinations = new HashMap<>();
+        listOfNaCombinations.forEach(na -> naCombinations.put(na, false));
```
Godelaine, line 346: could you try to make a unit test associated to the functionality you fixed ? 
and more complicated tests with naCombinations that doesn't have only false values

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTree.java
```diff
@@ -269,15 +276,23 @@ private void updateOptimalLeafWithNextDepthBestLeaf(AbstractNetworkPool networkP
...
+                        // TODO: we can wonder if it's better to do this here or at creation of each leaves or at each evaluation/optimization
+
+                        if (naCombinations.get(naCombination)) {
+                            previousDepthOptimalLeaf.getRangeActions().forEach(ra ->
+                                ra.apply(networkClone, input.getPrePerimeterResult().getRangeActionSetpointResult().getSetpoint(ra))
+                            );
+                        } else {
+                            input.getOptimizationPerimeter().getRangeActions()
+                                .forEach(ra ->
+                                        ra.apply(networkClone, previousDepthOptimalLeaf.getOptimizedSetpoint(ra, input.getOptimizationPerimeter().getMainOptimizationState()))
```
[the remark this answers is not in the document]
MartinBelthle: In this case, the boolean is false which means we want to keep RangeActions applied in the previousDepthOptimalLeaf. 

We want to have the setpoint corresponding to the result of this Leaf (previousDepthOptimalLeaf.getOptimizedSetpoint) and not the PrePerimeterResult one (what you wrote above)

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTree.java
```diff
@@ -246,9 +252,10 @@ private void iterateOnTree() {
      */
     private void updateOptimalLeafWithNextDepthBestLeaf(AbstractNetworkPool networkPool) throws InterruptedException {
 
-        final List<NetworkActionCombination> naCombinations = bloomer.bloom(optimalLeaf, input.getOptimizationPerimeter().getNetworkActions());
+        final Map<NetworkActionCombination, Boolean> naCombinations = bloomer.bloom(optimalLeaf, input.getOptimizationPerimeter().getNetworkActions());
```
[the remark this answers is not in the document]
MartinBelthle: I'm not sure we should do this because right after the code you highlighted we do this : 

` naCombinations.sort(this::deterministicNetworkActionCombinationComparison);`

It was done before our development to sort combinations a certain way. So I do not see why using a SortedMap could be useful if we sort its keys a different way immediately after.

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTree.java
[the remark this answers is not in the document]
Godelaine: sorry, i wrote the comment on the wrong line. I was referring to the point discussed yesterday for which you said there was a specific reason explaining why we couldn't use exactly the same input. .... as previously

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTree.java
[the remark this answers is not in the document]
Godelaine: yes, i meant couldn't you handle one object (a sorted map), rather than having to create a second object (the list ?) 
if you had a sorted map you could directly use the keyset

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTree.java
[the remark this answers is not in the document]
MartinBelthle: Oh yes I remember. In fact, the reason I thought we needed this change is not valid anymore :/. 

I can go back the previous behavior but i fail to understand the difference.

 Because before this commit you made in February https://github.com/farao-community/farao-core/pull/728/commits/67b8c668d8625f72cadc159d9ef8c327105ba488 inside your PR "Consider ARAs in 2P", it was  **previousDepthOptimalLeaf**.getRangeActions() and you changed it to **input.getPreOptimizationPerimeter**.getRangeActions()

ra-optimisation/search-tree-rao/src/test/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTreeBloomerTest.java
MartinBelthle: What do you mean by the functionality we fixed ? Do you mean a complete test that goes through SearchTree and removes RangeActions if the boolean is True ?

For the second part, we already implemented unit tests that have true values  &#x1F609;

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTreeBloomer.java
MartinBelthle: We did not find a more efficient way to do this. If you want we can extract a method from this one to make it lighter.

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/Leaf.java
MartinBelthle: I renamed it with its old name because it was more precise so i consider this conversation resolved

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTree.java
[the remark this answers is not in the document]
Godelaine: Thanks ! I don't rememeber the context behind this commit but could you check that nothing's broken on cases with a 2A ?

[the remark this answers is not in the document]
MartinBelthle: I ran every cucumber test tagged 2nd preventive (and also every "fast" tests) and they all pass, whether it's **previousDepthOptimalLeaf**.getRangeActions() or **input.getOptimizationPerimeter**.getRangeActions(). 

To make the code easier to read, i will choose **previousDepthOptimalLeaf** as it's also used in the `if` section right above.

phiedw (changes requested):

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTree.java
```diff
@@ -269,15 +277,23 @@ private void updateOptimalLeafWithNextDepthBestLeaf(AbstractNetworkPool networkP
...
-                        // Reset range action to their pre-perimeter set-points
-                        // TODO : instead of this parameter, clean rangeActions containted in Leaf. There should only be the optimized state's range actions
-                        // but then other RangeActionActivatinoResult in FARAO initialized with leaf's range actions should be initialized with prePerimeterOutput range Actions + leaf's
-                        // optimized range actions.
-                        input.getOptimizationPerimeter().getRangeActions().forEach(ra ->
-                            ra.apply(networkClone, input.getPrePerimeterResult().getRangeActionSetpointResult().getSetpoint(ra))
-                        );
-
-                        optimizeNextLeafAndUpdate(naCombination, networkClone);
+                        boolean shouldRangeActionBeRemoved = naCombinationsSorted.get(naCombination);
```
phiedw, line 280: Do you have a unit test that checks this new behaviour?

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTree.java
```diff
@@ -388,14 +404,15 @@ void optimizeNextLeafAndUpdate(NetworkActionCombination naCombination, Network n
...
+    Leaf createChildLeaf(Network network, NetworkActionCombination naCombination, boolean shouldRangeActionBeRemoved) {
         return new Leaf(
             input.getOptimizationPerimeter(),
             network,
             previousDepthOptimalLeaf.getActivatedNetworkActions(),
             naCombination,
+            shouldRangeActionBeRemoved ? new RangeActionActivationResultImpl(input.getPrePerimeterResult()) : previousDepthOptimalLeaf.getRangeActionActivationResult(),
             input.getPrePerimeterResult(),
-            input.getPreOptimizationAppliedRemedialActions());
+            shouldRangeActionBeRemoved ? input.getPreOptimizationAppliedRemedialActions() : getAppliedRemedialActions(previousDepthOptimalLeaf));
```
phiedw, line 415: add a unit test to check the new behaviour

phiedw (changes requested):

ra-optimisation/search-tree-rao/src/main/java/com/farao_community/farao/search_tree_rao/search_tree/algorithms/SearchTree.java
MartinBelthle: No but we do have a new cucumber test in the branch `Better handling of max RA`.
This test has a `max-mip-iterations` value of 1.
With master branch, the test fails because it cannot converge fast enough (because it "forgot" the PST setpoints of the RootLeaf.
On our cucumber branch, however, it can find a better solution because it remembers the SPT setpoints of the rootLeaf.

phiedw (approved):

PR #798, opened 2023-09-14
fix bug for zip file without subdirectory

jipea:

**Other information**:


data/crac-creation/crac-creator-csa-profiles/pom.xml
```diff
@@ -104,7 +103,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-inline</artifactId>
-            <scope>test</scope>
+            <scope>compile</scope>
```
benrejebmoh, line 106: do we really need mockito outside test scope?

data/crac-creation/crac-creator-csa-profiles/pom.xml
```diff
@@ -74,7 +74,6 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>farao-crac-impl</artifactId>
             <version>${project.version}</version>
-            <scope>runtime</scope>
```
benrejebmoh, line 77: i think it's better to keep runtime here,  you need to provide the implementation at the runtime in the built but you don't need to make that code accessible.

benrejebmoh (approved):

PR #839, opened 2023-11-21
INCLUDED combinationConstraintKinds -> AVAILABLE usageMethod

bqth29:

According to the ENTSO-E's specification, both `included` and `considered` combination constraint kinds make the remedial action available after a given contingency (there is actually no forcing the remedial action). This PR changed the UsageMethod of onContingencyState remedial actions from `FORCED` to `AVAILABLE` for all remedial actions. For ARAs, the UsageMethod as been set to `FORCED`.

**Other information**:


bqth29:
If this behavior is validated, it would be nice to check that the use of `included` leads to exclusive associations:
- a remedial action cannot be included for two distinct contingencies / assessed elements
- an assessed element with isCombinableWith... cannot have an AExCO/RA with included keyword

bqth29:
add for onConstraint

bqth29:
It was agreed that the current behaviour will be kept, i.e. CONSIDERED = AVAILABLE / INCLUDED = FORCED

PR #840, opened 2023-11-21
[CSA-79] Import FlowCNECs from AssessedElement defined with ConductingEquipment

bqth29:
AssessedElement defined with a ConductingEquipment instead of an OperationalLimit can be imported. The PATL and TATLs are queried directly in the network.

The mRID of the AssessedElement is now included in the CNEC name to ensure the uniqueness. The new format is `TSO_name (mRID) - contingency - instant`.

The CSA CNEC importer has been split in three different classes, one for each type of CNEC (Angle, Flow, Voltage).

Finally, in PowSyBl, when a line has no PATL, the TATL with the longest acceptableDuration is also used as the PATL. For coherence sake, this behaviour has also been implemented if the AssessedElement is defined using an OperationalLimit. If said limit is a TATL, it also used as a PATL if the attribute inBaseCase is true (this is logged in the creation context) to create a preventive version of the CNEC.


data/crac-creation/crac-creator-csa-profiles/src/main/java/com/farao_community/farao/data/crac_creation/creator/csa_profile/crac_creator/cnec/FlowCnecCreator.java
```diff
@@ -0,0 +1,321 @@
...
+    private final String conductingEquipment;
+    private final Set<Side> defaultMonitoredSides;
+
+    public FlowCnecCreator(Crac crac, Network network, String assessedElementId, String nativeAssessedElementName, String assessedElementOperator, boolean inBaseCase, PropertyBag currentLimitPropertyBag, String conductingEquipment, List<Contingency> linkedContingencies, Set<CsaProfileElementaryCreationContext> csaProfileCnecCreationContexts, CsaProfileCracCreationContext cracCreationContext, Set<Side> defaultMonitoredSides, String rejectedLinksAssessedElementContingency) {
+        super(crac, network, assessedElementId, nativeAssessedElementName, assessedElementOperator, inBaseCase, currentLimitPropertyBag, linkedContingencies, csaProfileCnecCreationContexts, cracCreationContext, rejectedLinksAssessedElementContingency);
+        this.conductingEquipment = conductingEquipment;
+        this.defaultMonitoredSides = defaultMonitoredSides;
+    }
+
+    public void addFlowCnecs() {
```
jipea, line 37: Beware of method names : there are addFlowCnecs, addFlowCnec, addFlowCnecData, addAllFlowCnecs in this class. I think you should find more explicit names, mainly for the last two.

data/crac-creation/crac-creator-csa-profiles/src/main/java/com/farao_community/farao/data/crac_creation/creator/csa_profile/crac_creator/cnec/FlowCnecCreator.java
```diff
@@ -0,0 +1,321 @@
...
+            return;
+        }
+
+        String networkElementId = definitionMode == FlowCnecDefinitionMode.CONDUCTING_EQUIPMENT ? conductingEquipment : operationalLimitPropertyBag.getId(CsaProfileConstants.REQUEST_OPERATIONAL_LIMIT_TERMINAL);
+        Identifiable<?> branch = getFlowCnecBranch(networkElementId);
+        if (branch == null) {
+            return;
+        }
+
+        Pair<EnumMap<Branch.Side, Double>, Map<Integer, EnumMap<Branch.Side, Double>>> thresholds = definitionMode == FlowCnecDefinitionMode.CONDUCTING_EQUIPMENT ? getPermanentAndTemporaryLimitsOfBranch((Branch<?>) branch) : getPermanentAndTemporaryLimitsOfOperationalLimit(branch, networkElementId);
```
jipea, line 49: In my opinion, this Pair type isn't very useful, it would be more readable if you use only the Map on the right of the pair, with the key Integer.MAX_VALUE for patl for example. You only have to extract this from the Map on line 50 and suppress the key from the map on line 51.
Alternatively, you can have a method for patl, and one for tatl, with definitionMode in argument, to cut algorithm in another way and avoid this Pair encapsulation.

jipea (changes requested):

jipea (approved):

PR #866, opened 2024-01-15
fix(optimization): add InitialScenario to avoid conflicts with Preventive one

MartinBelthle:

No

Bug fix

The preventive RAO's root leaf works on the "PReventiveScenario" variant of the network and doesn't roll-back changes (ie optimized range actions) before exiting the preventive RAO. If a range actions happens to be modified by the root leaf but put back to its original setpoint by the optimal leaf, this range action is not listed in the optimal leaf's applied RAs. Thus, after PRAO, the function that sets the optimal preventive setpoints for RAs skips this RA. Then the network has the root leaf's optimal setpoint set for this RA. This leads to a wrong network used in the whole curative RAO.
Example: 
1) Initially a PST is on tap "0".
2) The preventive root leaf modifies the tap to "10". When entering the first depth, it applies the new tap on the network.
3) Then, we find a better combination in first depth which modifies the tap to "0". We only apply it on a clone of the network (when parallalizing leaves).
4) As the tap rollbacked to the initial one, we do not consider it was activated in the searchTree (which is true) and therefore we do not apply the tap "0" on the network, which is a mistake.
5) The network being consider as the reference, for the following RAO steps we will consider the tap to be at position "10" instead of "0".

Now, the main algorithm resets the network to the state it was in before preventive RAO, right after preventive RAO (before applying optimal PRAs and running curative RAO)
Following the same example:
1) We introduce a network variant named "InitialScenario".
2) We follow the same steps until the preventive perimeter ends.
3) Once it is over, we use this network variant "InitialScenario" to apply activatedRemedialActions during preventive.
This way we avoid the described issue as the tap will still be "0" after preventive.

- [ ] Yes
- [x] No

**Other information**

An opened PR on farao-cucumber (branch `add-initial-network-scenario`) contains a test case similar to the one described above.
With this branch, all cucumber tests passed.


pet-mit (approved):

pet-mit (approved):
ok for me. CastorFullOpt & SearchTree deserve a refactorization IMO

PR #870, opened 2024-01-17
Fix divergence case

MartinBelthle:

AlgoLeague ticket 129.

Bug fixes

**Part 1 :**  If the RAO does a second preventive and fallbacks to the initial situation in the end, it encounters this error : 
`The RaoResult object should not be modified outside of its usual routine`.
 It's because its variable `optimizationStepsExecuted` went from `FIRST_PREVENTIVE_ONLY` to `SECOND_PREVENTIVE_FELLBACK_TO_INITIAL_SITUATION` which was not allowed.

**Part 2 :**
1) The method getVirtualCost() of the class SkippedOptimizationResultImpl returns 0.
2) The method getVirtualCost("sensitivity-failure-cost") returns 0.
2) When merging preventive and curative perimeter, we filter out contingencies with a load flow divergence.

**Example:**

CASTOR applies PRAs. These actions make a contingency perimeter "Co-1 curative" diverge. 
So, it skips this perimeter and marks it as a SkippedOptimizationResultImpl.
When merging preventive and curative perimeters, it filters out this perimeter from the cost evaluation.
This means even though CASTOR made a perimeter diverge, in the end it does not see it.
To correct this behavior, we also have to change the value returned by the 2 methods `getVirtualCost` as even if CASTOR considered this perimeter it would still not see the virtual cost associated.

**Part 1 :** The exception no longer occurs as this change is allowed.

**Part 2 :**
1) The class SkippedOptimizationResultImpl now has a variable `sensitivityFailureOvercost` which corresponds to the one filled in the RAO Parameters. The method getVirtualCost() now returns this cost. The same goes for getVirtualCost(virtualCostName) which returns  `sensitivityFailureOvercost` if called with "sensitivity-failure-cost" and 0 otherwise.
2) When merging preventive and curative perimeter, we do not longer filter out contingencies with a load flow divergence.

So, if the same example occurs, the RAO will go through every contingency perimeter and for the ones with loadflow divergence, it will add the sensitivityFailureOvercost to the virtual cost.

- [ ] Yes
- [x] No

**Other information**:

The farao-cucumber-branch `swe-divergence` contains a test case and every test passed on Jenkins.


ra-optimisation/rao-api/src/main/java/com/powsybl/openrao/raoapi/parameters/LoadFlowAndSensitivityParameters.java
```diff
@@ -58,7 +59,10 @@ public double getSensitivityFailureOvercost() {
     }
 
     public void setSensitivityFailureOvercost(double sensitivityFailureOvercost) {
-        this.sensitivityFailureOvercost = sensitivityFailureOvercost;
+        if (sensitivityFailureOvercost < 0) {
+            BUSINESS_WARNS.warn("The value {} for `sensitivity-failure-overcost` is smaller than 0. This would encourage the optimizer to make the loadflow diverge. Thus, it will be set to + {}", sensitivityFailureOvercost, -sensitivityFailureOvercost);
```
pet-mit, line 63: throw exception instead (we do it elsewhere when the parameter is wrong, for example in setPstSensitivityThreshold)

pet-mit (changes requested):

ra-optimisation/rao-api/src/main/java/com/powsybl/openrao/raoapi/parameters/LoadFlowAndSensitivityParameters.java
MartinBelthle: I talked about it with Philippe and we copied the logic of the method `setCurativeMinObjImprovement` in class `ObjectiveFunctionParameters`. I can modify the code but I prefer that you discuss it with him before.

pet-mit: ok then, for now. but i think we should harmonize the parameters more, and see what is done in other powsybl projects

pet-mit (approved):

PR #886, opened 2024-02-13
feat(parameters): add ra usage limits per instant

MartinBelthle:

No.

New feature

It's only possible to specify RAs usage limits for curative instant.

It's now possible to do so for other instants.

- [x] Yes
- [ ] No

**Before :**

RaoParameters v2.1 contained this section : 
```
  "ra-usage-limits-per-contingency" : {
    "max-curative-ra" : 2147483647,
    "max-curative-tso" : 1,
    "max-curative-topo-per-tso" : { },
    "max-curative-pst-per-tso" : {"be" : 0},
    "max-curative-ra-per-tso" : { }
  }
```

**Now:**

RaoParameters should be in version 2.2. Also the previous section does not exist anymore.
Instead, to reproduce the same behavior as the one in the example above, you should add the following section in the crac Creation Parameters

```
  "ra-usage-limits-per-instant": [ {
    "instant": "curative",
    "max-ra" : 2147483647,
    "max-tso" : 1,
    "max-topo-per-tso" : { },
    "max-pst-per-tso" : {"be" : 0},
    "max-ra-per-tso" : { }
  } ]
```

The field `instant` in the previous example can take any value of an instant registered in the CRAC and the RAs limits will be applied to the given instant. See a complete example below : 

```
  "ra-usage-limits-per-instant": [
{
    "instant": "preventive",
    "max-ra" : 3,
    "max-tso" : 6,
    "max-topo-per-tso" : {"fr":4, "be": 2},
    "max-pst-per-tso" : {"fr" : 6},
    "max-ra-per-tso" : {"fr": 12}
  },
 {
    "instant": "curative",
    "max-ra" : 2147483647,
    "max-tso" : 1,
    "max-topo-per-tso" : { },
    "max-pst-per-tso" : {"be" : 0},
    "max-ra-per-tso" : { }
  } ]

```

**Other information**:
All tests passed on cucumber branch `feat/add-ra-usage-limits-per-instant` which contains added tests.


data/crac-io/crac-io-json/src/main/java/com/powsybl/openrao/data/craciojson/deserializers/RaUsageLimitsDeserializer.java
```diff
@@ -0,0 +1,28 @@
+package com.powsybl.openrao.data.craciojson.deserializers;
```
pet-mit, line 1: header & author

data/crac/crac-api/src/main/java/com/powsybl/openrao/data/cracapi/Crac.java
```diff
@@ -510,4 +510,14 @@ default State getState(String contingencyId, Instant instant) {
      * Remove a NetworkAction - identified by its id - from the Crac
      */
     void removeNetworkAction(String id);
+
+    /**
+     * Get all remedial action usage limitation according to the crac creation parameters
+     */
+    Map<String, RaUsageLimits> getRaUsageLimitsPerInstant();
```
pet-mit, line 517: once "inside" CRAC and RAO, there is no reason to keep IDs instead of Instants
(this applies for all usages inside RAO too)
```suggestion
    Map<Instant, RaUsageLimits> getRaUsageLimitsPerInstant();
```
also, add a getRaUsageLimits(Instant instant) method

data/crac/crac-api/src/main/java/com/powsybl/openrao/data/cracapi/RaUsageLimits.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private static void crossCheckMaxCraPerTsoParameters(Map<String, Integer> maxRaPerTso, Map<String, Integer> maxTopoPerTso, Map<String, Integer> maxPstPerTso) {
+        Set<String> tsos = new HashSet<>();
+        tsos.addAll(maxRaPerTso.keySet());
+        tsos.addAll(maxTopoPerTso.keySet());
+        tsos.addAll(maxPstPerTso.keySet());
+        tsos.forEach(tso -> {
+            if (maxTopoPerTso.containsKey(tso)
+                && maxRaPerTso.getOrDefault(tso, 1000) < maxTopoPerTso.get(tso)) {
```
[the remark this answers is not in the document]
MartinBelthle: I simply copied what was done in the RaoParameters but indeed it's not very reliable, i've fixed it.

pet-mit (commented):
I wonder if we mustn't create adders for the new limits in the CRAC
crac.newRaUsageLimits("preventive")
       .withMaxTso(...)
       ...
       .add()

data/crac/crac-api/src/main/java/com/powsybl/openrao/data/cracapi/Crac.java
MartinBelthle: To achieve this, i should at some point convert String (inside CracCreationParameters) to Instant (inside Crac).
So i should check the string value to infer its InstantKind.
Is this what you had in mind ?

data/crac-io/crac-io-json/pom.xml
```diff
@@ -75,6 +75,12 @@
             <artifactId>powsybl-commons-test</artifactId>
             <version>${powsybl.core.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.powsybl</groupId>
+            <artifactId>open-rao-crac-creator-api</artifactId>
+            <version>5.2.0-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
```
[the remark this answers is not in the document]
pet-mit: what you are trying to do here is to create two files that can read and write RaUsageLimits (CRAC creation parameters & CRAC). It just happens that both are in json format, and both are written the same way, in the current version.
In order to avoid code duplication, you created a depedency which should not exist.
I'd rather have the code duplication, or the common code moved to a third "utils" package

data/crac/crac-api/src/main/java/com/powsybl/openrao/data/cracapi/Crac.java
pet-mit: you don't need to deduce the instantKind, only the instant. you can do it inside every crac creator, or in the setter inside CracImpl : when the CRAC recieves the map with string keys, it maps and stores the strings to the instant objects it carries. But I do think that it would be cleaner to create an adder, like for all other CRAC objects

data/crac-io/crac-io-json/pom.xml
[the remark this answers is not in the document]
MartinBelthle: I understand.
I would like to avoid duplicating the code and rather moving it into the package `crac-creation-util`. 
The only issue is that doing so creates a circual dependency for only 1 reason : `OpenRaoImportException` needs `ImportStatus` which is located here `data.craccreation.creator.api` :disappointed: .
Moving `ImportStatus`to the `crac-creation-util` would solve the issue and avoid code duplication.
But it would mean this PR modifies tons of files. 
Do you agree with this solution or should I just give up and duplicate the code?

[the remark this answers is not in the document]
pet-mit: just duplicate the code since, from a functional point of view, there is no reason to always have the same

data/crac-creation/crac-creator-csa-profiles/src/main/java/com/powsybl/openrao/data/craccreation/creator/csaprofile/craccreator/CsaProfileCracCreator.java
```diff
@@ -89,6 +90,17 @@ private void addCsaInstants() {
...
+    private void addRaUsageLimits(CracCreationParameters parameters) {
+        parameters.getRaUsageLimitsPerInstant().forEach((instantName, raUsageLimits)
+            -> crac.newRaUsageLimits(instantName)
+            .withMaxRa(raUsageLimits.getMaxRa())
+            .withMaxTso(raUsageLimits.getMaxTso())
+            .withMaxRaPerTso(raUsageLimits.getMaxRaPerTso())
+            .withMaxPstPerTso(raUsageLimits.getMaxPstPerTso())
+            .withMaxTopoPerTso(raUsageLimits.getMaxTopoPerTso())
+            .add());
+    }
```
pet-mit, line 102: put this in a new common class in crac-creation-util, since it is used by all CRAC creators

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/RaUsageLimitsAdderImpl.java
```diff
@@ -0,0 +1,63 @@
+package com.powsybl.openrao.data.cracimpl;
```
pet-mit, line 1: header & author

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/RaUsageLimitsAdderImpl.java
pet-mit, line 1: add a test class for this

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/searchtree/parameters/SearchTreeParameters.java
```diff
@@ -104,6 +111,29 @@ public int getMaxNumberOfIterations() {
...
+            String tso = rangeAction.getOperator();
+            tsoCount.add(tso);
+            raCount += 1;
+            currentRaPerTsoLimits.computeIfPresent(tso, (key, currentLimit) -> Math.max(0, currentLimit - 1));
+            currentPstPerTsoLimits.computeIfPresent(tso, (key, currentLimit) -> Math.max(0, currentLimit - 1));
+        }
+        raUsageLimits.setMaxRa(Math.max(0, raUsageLimits.getMaxRa() - raCount));
+        raUsageLimits.setMaxTso(Math.max(0, raUsageLimits.getMaxTso() - tsoCount.size()));
+        currentTopoPerTsoLimits.forEach((tso, raLimits) -> currentTopoPerTsoLimits.put(tso, Math.min(raLimits, currentRaPerTsoLimits.getOrDefault(tso, 1000))));
+        currentPstPerTsoLimits.forEach((tso, raLimits) -> currentPstPerTsoLimits.put(tso, Math.min(raLimits, currentRaPerTsoLimits.getOrDefault(tso, 1000))));
```
pet-mit, line 130: change "1000"

pet-mit (changes requested):

data/crac-creation/crac-creator-csa-profiles/src/main/java/com/powsybl/openrao/data/craccreation/creator/csaprofile/craccreator/CsaProfileCracCreator.java
MartinBelthle: I realize that I named it RaUsageLimitsAdder as it's the only thing it does but it's also the name of an Interface in crac api package. 
Perhaps it should be renamed otherwise to avoid misunderstandings.

data/crac/crac-impl/src/test/java/com/powsybl/openrao/data/cracimpl/RaUsageLimitsAdderImplTest.java
```diff
@@ -0,0 +1,75 @@
...
+        assertEquals(new RaUsageLimits(), crac.getRaUsageLimits(outageInstant));
+        assertEquals(new RaUsageLimits(), crac.getRaUsageLimits(curativeInstant));
+
+        // with empty limitations
+        crac.newRaUsageLimits("preventive").add();
+        assertEquals(new RaUsageLimits(), crac.getRaUsageLimits(preventiveInstant));
+
+        // with fake instant
+        RaUsageLimits nullRaUsageLimits = crac.newRaUsageLimits("fakeInstant").add();
+        assertNull(nullRaUsageLimits);
```
pet-mit, line 54: this should throw an exception

data/crac/crac-impl/src/test/java/com/powsybl/openrao/data/cracimpl/utils/ExhaustiveCracCreation.java
```diff
@@ -23,10 +23,10 @@
...
  */
 public final class ExhaustiveCracCreation {
-    private static final String PREVENTIVE_INSTANT_ID = "preventive";
-    private static final String OUTAGE_INSTANT_ID = "outage";
-    private static final String AUTO_INSTANT_ID = "auto";
-    private static final String CURATIVE_INSTANT_ID = "curative";
+    public static final String PREVENTIVE_INSTANT_ID = "preventive";
+    public static final String OUTAGE_INSTANT_ID = "outage";
+    public static final String AUTO_INSTANT_ID = "auto";
+    public static final String CURATIVE_INSTANT_ID = "curative";
```
pet-mit, line 29: I don't love these changes ; these IDs are supposed to be used with unit tests that use the CRAC created by this class only

pet-mit (changes requested):

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/RaUsageLimitsAdderImpl.java
```diff
@@ -7,31 +7,27 @@
...
 
     RaUsageLimitsAdderImpl(CracImpl owner, String instantName) {
         Objects.requireNonNull(owner);
         this.owner = owner;
         List<Instant> instants = this.owner.getSortedInstants().stream().filter(cracInstant -> cracInstant.getId().equals(instantName)).toList();
         if (instants.isEmpty()) {
-            BUSINESS_WARNS.warn("The instant {} registered in the crac creation parameters does not exist in the crac. Its remedial action limitations will be ignored.", instantName);
-        } else {
-            this.instant = instants.get(0);
+            throw new OpenRaoException(String.format("The instant %s registered in the crac creation parameters does not exist in the crac. Its remedial action limitations will be ignored.", instantName));
```
pet-mit, line 28: you dont't know that the user is using a crac creation parameters, maybe they are calling the adder API. plus, the RA limitations will not only be ignored, since you threw an exception, so the user's app will crash.
```suggestion
            throw new OpenRaoException(String.format("The instant %s does not exist in the crac.", instantName));
```

pet-mit (changes requested):

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/RaUsageLimitsAdderImpl.java
MartinBelthle: Indeed, i committed to quickly.

pet-mit (changes requested):
OK for me but it seems you're mising new tests on the CRAC json IO modules. add a new round-trip test on the new version, containing all RA usage limits

MartinBelthle:
I've added a test that uses `ExhaustiveCracCreation`

pet-mit (approved):

PR #912, opened 2024-03-05
Network action compatibility filter

bqth29:

In the search tree, several network can be incompatible so if one was applied, it is useless to try to apply incompatible remedial actions because they would just override the currently applied remedial action's behaviour.

Two elementary actions are compatible if:
- they have different types (ex: topo and injection)
- they have the same type but different network elements
- they have the same type and the same network elements + the same action type for topo / the same setpoint for injections

A network action NA1 is said to be included in another network action NA2 (NA1⊂NA2) if all the elementary actions of NA1 are elementary actions of NA2.

NA1 and NA2 are equal if all their elementary actions are equal (NA1⊂NA2 and NA2⊂NA1)

Two network actions are compatible if :

- one is not included in the other
- (thus they are different)
- their elementary actions are compatible pairwise

otherwise they are incompatible.

A list of network actions is incompatible if at least two of there network actions are not compatible.


ra-optimisation/search-tree-rao/src/test/java/com/powsybl/openrao/searchtreerao/searchtree/algorithms/SearchTreeBloomerTest.java
```diff
@@ -569,4 +574,88 @@ void testDontFilterNullOperator() {
 
         assertEquals(1, filteredNaCombination.size()); // no combination filtered, because null operator should not count
     }
+
+    @Test
+    void removeIncompatibleCombinations() {
```
pet-mit, line 579: replace test setup with an actual CRAC creation, it would be easier to read & maintain than mocks

util/src/main/java/com/powsybl/openrao/util/NetworkActionsCompatibilityChecker.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @author Thomas Bouquet {@literal <thomas.bouquet at rte-france.com>}
+ */
+public final class NetworkActionsCompatibilityChecker {
+    private NetworkActionsCompatibilityChecker() { }
+
+    public static boolean areNetworkActionsCompatible(NetworkAction networkAction1, NetworkAction networkAction2) {
```
pet-mit, line 29: declare a new API function instead : networkAction.isCompatibleWith(otherNetworkAction)

util/src/main/java/com/powsybl/openrao/util/NetworkActionsCompatibilityChecker.java
```diff
@@ -0,0 +1,107 @@
...
+            for (ElementaryAction elementaryAction2 : networkAction2.getElementaryActions()) {
+                if (!areElementaryActionsCompatible(elementaryAction1, elementaryAction2)) {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
+
+    static boolean areElementaryActionsCompatible(ElementaryAction elementaryAction1, ElementaryAction elementaryAction2) {
```
pet-mit, line 40: declare a new API function instead : elementaryAction.isCompatibleWith(otherElementaryAction)

util/src/main/java/com/powsybl/openrao/util/NetworkActionsCompatibilityChecker.java
```diff
@@ -0,0 +1,107 @@
...
+            return areSwitchPairsCompatible((SwitchPair) elementaryAction1, (SwitchPair) elementaryAction2);
+        } else {
+            throw new OpenRaoException("Unsupported network action type: " + elementaryAction1.getClass().getName());
+        }
+    }
+
+    private static boolean areTopologicalActionAndSwitchPairCompatible(TopologicalAction topologicalAction, SwitchPair switchPair) {
+        return ActionType.OPEN.equals(topologicalAction.getActionType()) && topologicalAction.getNetworkElement().equals(switchPair.getSwitchToOpen())
+            || ActionType.CLOSE.equals(topologicalAction.getActionType()) && topologicalAction.getNetworkElement().equals(switchPair.getSwitchToClose());
+    }
```
pet-mit, line 63: forget this for now

pet-mit (changes requested):

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/NetworkActionImpl.java
```diff
@@ -58,6 +58,18 @@ public Set<NetworkElement> getNetworkElements() {
...
+    public boolean isCompatibleWith(NetworkAction otherNetworkAction) {
+        for (ElementaryAction elementaryAction1 : elementaryActions) {
+            for (ElementaryAction elementaryAction2 : otherNetworkAction.getElementaryActions()) {
+                if (!elementaryAction1.isCompatibleWith(elementaryAction2)) {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
```
pet-mit, line 71: declare this as default behaviour in NetworkAction interface

util/src/main/java/com/powsybl/openrao/util/NetworkActionsCompatibilityChecker.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.powsybl.openrao.data.cracapi.networkaction.NetworkAction;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * @author Thomas Bouquet {@literal <thomas.bouquet at rte-france.com>}
+ */
+public final class NetworkActionsCompatibilityChecker {
```
pet-mit, line 18: no need for this class, put its code in the bloomer

pet-mit (changes requested):

pet-mit (approved):

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/SwitchPairImpl.java
```diff
@@ -57,6 +58,15 @@ public NetworkElement getSwitchToClose() {
         return switchToClose;
     }
 
+    @Override
+    public boolean isCompatibleWith(ElementaryAction otherElementaryAction) {
+        if (otherElementaryAction instanceof SwitchPair switchPair) {
+            return !switchToOpen.equals(switchPair.getSwitchToOpen()) && !switchToOpen.equals(switchPair.getSwitchToClose()) && !switchToClose.equals(switchPair.getSwitchToClose()) && !switchToClose.equals(switchPair.getSwitchToOpen())
+                || switchToOpen.equals(switchPair.getSwitchToOpen()) || switchToClose.equals(switchPair.getSwitchToClose());
```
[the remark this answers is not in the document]
phiedw: (or simply was the second || supposed to be &&?)

phiedw (changes requested):

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/SwitchPairImpl.java
```diff
@@ -57,6 +58,14 @@ public NetworkElement getSwitchToClose() {
         return switchToClose;
     }
 
+    @Override
+    public boolean isCompatibleWith(ElementaryAction otherElementaryAction) {
+        if (otherElementaryAction instanceof SwitchPair switchPair) {
+            return !switchToOpen.equals(switchPair.getSwitchToOpen()) && !switchToOpen.equals(switchPair.getSwitchToClose()) && !switchToClose.equals(switchPair.getSwitchToClose()) && !switchToClose.equals(switchPair.getSwitchToOpen());
```
pet-mit, line 64: I'm not sure this is OK. We may have a bus bar change that changes bus from 1 to 2, another from 2 to 3. combining both allows the RAO to move from 1 to 3, but this method wouldn't allow it anymore.
what do you think @phiedw ?

pet-mit: I'm thinking that "icompatible" RAs should be viewed as one RA that completely overrides the effect of the other, not just modifies part of its effect

phiedw: I think we should ask Sebastien what the TSOs want in this case in particular, the approach here is the most "safe" but perhaps indeed they allow multiple bus bar changes for the same line (but imo they would define the switch from 1 to 3 directly if that were the case)

phiedw: This is currently only used to go from 1 node topo to 2 node topo and other way round, so the case where we would move a line from bus bar 1 to 2, and then to 3 can probably be ignored for now (since it would also require some changes in other parts of the RAO to make sure things go smoothly, eg we currently just apply all network actions selected in some parts of the code and we would need to make sure we applied in the correct order for this case).

So I think for now this version is good.

pet-mit (approved):

phiedw (approved):

PR #920, opened 2024-03-11
Replace rao elementary action by core action

pjeanmarie:

- [ ] Yes
- [ ] No

**Other information**:


pjeanmarie:
Do not merge: Need next core version (6.4.0), then pom need to be updated (6.4.0-SNAPSHOT -> 6.4.0) and changes on CI files need to be removed.

pjeanmarie:
@pet-mit @obrix There were some bugs in the actual version not previously revealed by tests:
These bugs are in the FlowBasedConstraint CRAC format conversion to crac:
Contrary to what was done for all other formats (CSE, CIM, CSA), the type of the network element was not checked to create the appropriate action and raise error, the action created was always a terminal connection action assuming the type must always be a connectable which is not the case. This create 2 cases of incorrect actions (apply will raise exception), one for switch and one for tie-line. Fixes:
- if the type is a switch we can create a SwitchAction instead
- if the type is a TieLine, it is strange that the TieLine is not a Connectable, it is the only Branch which is not a Connectable, it has terminals. There is no Action or Modification for a TieLine but a TieLine is composed of two DanglingLines which are Connectable. So fix in rao (without fix in core) can be to create 2 terminals connection actions so this add one elementary actions but also its chanage the crac network elements ! Change in core could be to say that a TieLine is a Connectable or to modify TerminalsConnectionAction.toModification() to manage the TieLine case (and other network element with terminals but not in Connectable if any...).
NB. To check for other TerminalConnectionAction creation in CSE and CIM, if theorically, we can have a TieLine and not just Connectable...

annetill:
About issue with terminals connection action: the to modification is not able to handle tie line and hvdc line easily, so it won't be in this release. For this open-rao version, a tie line action must be translated in one or two dangling line actions. Same for hvdc lines. @pjeanmarie Can you show me where the fix is in your code?

pjeanmarie:
> About issue with terminals connection action: the to modification is not able to handle tie line and hvdc line easily, so it won't be in this release. For this open-rao version, a tie line action must be translated in one or two dangling line actions. Same for hvdc lines. @pjeanmarie Can you show me where the fix is in your code?

There is already a fix which do the action on the dangling lines (commit before last commit) in case of Tie Line: https://github.com/powsybl/powsybl-open-rao/pull/920/commits/508916c4fe247262222f43bdfed31a3732b4cd13

This fix is simple but we do not have the same network element in crac, we have the 2 dangling lines instead of the tie line. If it is a problem we need to change more code (new adder...).

NB. For HDVC line, can we have a hvdc line from a flow-based-constraint CRAC format? and from CIM, CSE, CSA CRAC format?

data/crac/crac-api/src/main/java/com/powsybl/openrao/data/cracapi/networkaction/DanglingLineActionAdder.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.openrao.data.cracapi.networkaction;
```
obrix, line 7: Here and for all adders, update copyright and author.

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/AbstractSingleNetworkElementActionAdderImpl.java
```diff
@@ -0,0 +1,42 @@
+package com.powsybl.openrao.data.cracimpl;
```
obrix, line 1: missing header and fix author

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/GeneratorActionAdderImpl.java
```diff
@@ -0,0 +1,47 @@
...
+    public GeneratorActionAdder withActivePowerValue(double activePowerValue) {
+        this.activePowerValue = activePowerValue;
+        return this;
+    }
+
+    protected Action buildAction() {
+        return new GeneratorActionBuilder()
+            .withId(String.format("%s_%s_%s", getActionName(), networkElementId, activePowerValue))
+            .withNetworkElementId(networkElementId)
+            .withActivePowerRelativeValue(false)
```
obrix, line 35: No possibility / need to have a relative value change action ?

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/LoadActionAdderImpl.java
```diff
@@ -0,0 +1,48 @@
...
+    public LoadActionAdder withActivePowerValue(double activePowerValue) {
+        this.activePowerValue = activePowerValue;
+        return this;
+    }
+
+    protected Action buildAction() {
+        return new LoadActionBuilder()
+            .withId(String.format("%s_%s_%s", getActionName(), networkElementId, activePowerValue))
+            .withNetworkElementId(networkElementId)
+            .withRelativeValue(false)
```
obrix, line 36: Same question for relative change ?

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/PhaseTapChangerTapPositionActionAdderImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        this.tapPosition = tapPosition;
+        return this;
+    }
+
+    protected Action buildAction() {
+        return new PhaseTapChangerTapPositionActionBuilder()
+            .withId(String.format("%s_%s_%s", getActionName(), networkElementId, tapPosition))
+            .withNetworkElementId(networkElementId)
+            .withTapPosition(tapPosition)
+            .withRelativeValue(false)
```
obrix, line 37: Same question for relative change

data/crac-io/crac-io-json/src/main/java/com/powsybl/openrao/data/craciojson/serializers/ShuntCompensatorPositionActionSerializer.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @author Peter Mitri {@literal <peter.mitri at rte-france.com>}
+ */
+public class ShuntCompensatorPositionActionSerializer extends AbstractJsonSerializer<ShuntCompensatorPositionAction> {
+    @Override
+    public void serialize(ShuntCompensatorPositionAction value, JsonGenerator gen, SerializerProvider serializers) throws IOException {
+        gen.writeStartObject();
+        gen.writeStringField(NETWORK_ELEMENT_ID, value.getShuntCompensatorId());
+        gen.writeNumberField(SETPOINT, value.getSectionCount());
```
obrix, line 27: SECTIONCOUNT would more relevant here than SETPOIN, don't know if it could be an issue though.

obrix (dismissed):
First quick review, it would be great to have a working CI soon.

data/crac/crac-api/src/main/java/com/powsybl/openrao/data/cracapi/networkaction/DanglingLineActionAdder.java
pjeanmarie: Done for all new files

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/AbstractSingleNetworkElementActionAdderImpl.java
pjeanmarie: Done for all new files

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/GeneratorActionAdderImpl.java
pjeanmarie: For this Issue, it was only ask to keep the same functionality than before, no more.
If we want more, it will need a lot of changes: updates adder, tests, json... it needs a separate PR since this one is already huge. NB. There is a lot more possibilities for a GeneratorAction than what Rao used here.

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/LoadActionAdderImpl.java
pjeanmarie: see previous answer

data/crac-io/crac-io-json/src/main/java/com/powsybl/openrao/data/craciojson/serializers/ShuntCompensatorPositionActionSerializer.java
pjeanmarie: This PR keep the json exactly as before (non-regression). For new json, and so new vocabulary, see: https://github.com/powsybl/powsybl-open-rao/pull/1034 (its in another PR to since this one is already huge).

pjeanmarie:
> First quick review, it would be great to have a working CI soon.

Now the CI works with 2 commits that should be removed: the first one which update pom to core 6.4.0 and the second one (CompareTxt no longer exists in powsybl-core) which fix Rao for this new core.
This PR do not need more, but there is a bump branch with update of all dependencies (not just core), but it seems to not work for now: https://github.com/powsybl/powsybl-open-rao/pull/1071

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/NetworkActionImpl.java
```diff
@@ -7,55 +7,116 @@
...
+            } else if (elementaryAction instanceof SwitchPair switchPair) {
+                return !network.getSwitch(switchPair.getSwitchToOpen().getId()).isOpen() || network.getSwitch(switchPair.getSwitchToClose().getId()).isOpen();
+            } else if (elementaryAction instanceof TerminalsConnectionAction terminalsConnectionAction) {
+                Connectable<?> connectable = network.getConnectable(terminalsConnectionAction.getElementId());
+                if (terminalsConnectionAction.isOpen()) {
+                    // Connectable is considered closed if all terminals are connected
+                    return connectable.getTerminals().stream().allMatch(Terminal::isConnected);
+                } else {
+                    // Connectable is already considered opened if one of the terminals is disconnected
+                    return !connectable.getTerminals().stream().allMatch(Terminal::isConnected);
```
annetill, line 72: To be checked by @pet-mit

data/crac-creation/crac-creation-util/src/main/java/com/powsybl/openrao/data/craccreation/util/ucte/UcteTopologicalElementHelper.java
```diff
@@ -69,6 +71,13 @@ public UcteTopologicalElementHelper(String ucteBranchId, UcteNetworkAnalyzer uct
         }
     }
 
+    /**
+     * If the connectable is valid, returns its type in the iidm network
+     */
+    public Identifiable<?> getIidmIdentifiable() {
```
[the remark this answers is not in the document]
pjeanmarie: There is another interface in Rao named Identifiable (in crac-api) + it came from already existing UcteMatchingResult getIidmIdentifiable()

data/crac-creation/crac-creator-cim/src/main/java/com/powsybl/openrao/data/craccreation/creator/cim/craccreator/NetworkActionCreator.java
```diff
@@ -207,14 +228,31 @@ private void addTopologicalElementaryAction(String elementaryActionId, RemedialA
             networkElementId = branchHelper.getIdInNetwork();
             element = branchHelper.getBranch();
         }
-        if (!(element instanceof Branch) && !(element instanceof Switch)) {
-            throw new OpenRaoImportException(ImportStatus.INCONSISTENCY_IN_DATA, String.format("%s is nor a branch nor a switch on elementary action %s", networkElementId, elementaryActionId));
+        if (element instanceof Connectable) {
```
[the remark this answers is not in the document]
pjeanmarie: To fix: the test should not be on Connectable but on Branch (more specific + tieline). I will check other Connectable test elsewhere that probably need the same kind of fix.

data/crac-creation/crac-creator-csa-profiles/src/main/java/com/powsybl/openrao/data/craccreation/creator/csaprofile/craccreator/remedialaction/CsaProfileRemedialActionsCreator.java
```diff
@@ -316,28 +318,29 @@ private Set<String> createRemedialActionGroups() {
...
-        topologicalActions.forEach(ea -> {
-            TopologicalAction topologicalAction = (TopologicalAction) ea;
-            networkActionAdder.newTopologicalAction()
-                .withNetworkElement(topologicalAction.getNetworkElement().getId())
-                .withActionType(topologicalAction.getActionType())
-                .add();
+    private static void addElementaryActionsToGroup(List<Action> elementaryActions, NetworkActionAdder networkActionAdder) {
+        elementaryActions.forEach(ea -> {
+            if (ea instanceof GeneratorAction generatorAction) {
+                networkActionAdder.newGeneratorAction()
```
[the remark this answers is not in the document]
pjeanmarie: Previously, ElementaryAction in Rao did not have an Id so I have nothing to fill the Action Id, so the rao Adder do not ask for id (iso functionality) but it generates one even if not used in Rao.
The id is still generated in a way to try to be unique, even if not use in Rao at the moment, to do that it is a combination of the action type, the network element id and the values of each parameters (one or 2 at most in Rao).

data/crac-creation/crac-creator-cse/src/main/java/com/powsybl/openrao/data/craccreation/creator/cse/remedialaction/TRemedialActionAdder.java
```diff
@@ -111,10 +108,20 @@ private void importTopologicalAction(TRemedialAction tRemedialAction) {
...
-                .withActionType(convertActionType(tBranch.getStatus()))
-                .add();
+            Identifiable<?> ne = network.getIdentifiable(branchHelper.getIdInNetwork());
+            if (ne.getType() == IdentifiableType.SWITCH) {
+                networkActionAdder.newSwitchAction()
+                    .withNetworkElement(branchHelper.getIdInNetwork())
+                    .withActionType(convertActionType(tBranch.getStatus()))
+                    .add();
+            } else if (ne instanceof Connectable<?>) {
+                networkActionAdder.newTerminalsConnectionAction()
```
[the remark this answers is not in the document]
pjeanmarie: Its iso-functionality to what the Rao was doing (all adder are iso-functionality): it was not possible to select a side in rao, the same modification (open/close) was executed on both terminals, its also the case now. We can discuss how to access more functionality of Action in the Rao but it is a separate Issue and in a technical point of view, we need a new PR, this one is already too big.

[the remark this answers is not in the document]
pjeanmarie: I will fix the Connectable with Branch

pjeanmarie:
@annetill @pet-mit
A new method, called isCompatibleWith, was added to the interface of NetworkAction and all the interfaces (4 before this dev) which inherit of ElementaryAction. The implementation/code of all isCompatible methods was not done in Impl class but direclty in the interfaces (one for NetworkAction which used the interface ElementaryAction and one for each interface which inherit ElementaryAction).
There is reasons for this default implem in interface but I am not sure: interface user should have this implem (and should no reimplement it?), tests with sonar... ?
Consequence: With the change ElementaryAction -> Action, and the default implem of isCompatibleWith to be done for all ElementaryAction, a default implem was necessary for all Action (which could be done on all its sub interface). We do not have yet the default implem of  isCompatibleWith in core, usually when its the case we used (temporarily) an implem at rao NetworkAction level using "instance of" (ex. hasImpactOnNetwork, canBeApplied) and we raised ExceptionNotImplemented for Action not used in Rao. Here we could do the same (but at interface level), covering only the small percent of Action used in Rao. But to have a simplier/smaller isCompatibleWith default implem and also to better cover Action interfaces, we decide to consider that for all Action, the compatibility can be checked by verifying that both actions are not of same type, or not on same network element, or, if they are, they are exaclty the same (NB. this is inderect and so not necessarily true, what should really be checked its if is they can change the same element of a piece of network in not the same way).
To do this simple default implem we were missing a getNetworkElement in Action interface and an implem of equals (using all attributs) for all Action (not only all Action used in rao). For now, without the FIXEDME, it works for all rao Action.

pet-mit:
> @annetill @pet-mit A new method, called isCompatibleWith, was added to the interface of NetworkAction and all the interfaces (4 before this dev) which inherit of ElementaryAction. The implementation/code of all isCompatible methods was not done in Impl class but direclty in the interfaces (one for NetworkAction which used the interface ElementaryAction and one for each interface which inherit ElementaryAction). There is reasons for this default implem in interface but I am not sure: interface user should have this implem (and should no reimplement it?), tests with sonar... ? Consequence: With the change ElementaryAction -> Action, and the default implem of isCompatibleWith to be done for all ElementaryAction, a default implem was necessary for all Action (which could be done on all its sub interface). We do not have yet the default implem of isCompatibleWith in core, usually when its the case we used (temporarily) an implem at rao NetworkAction level using "instance of" (ex. hasImpactOnNetwork, canBeApplied) and we raised ExceptionNotImplemented for Action not used in Rao. Here we could do the same (but at interface level), covering only the small percent of Action used in Rao. But to have a simplier/smaller isCompatibleWith default implem and also to better cover Action interfaces, we decide to consider that for all Action, the compatibility can be checked by verifying that both actions are not of same type, or not on same network element, or, if they are, they are exaclty the same (NB. this is inderect and so not necessarily true, what should really be checked its if is they can change the same element of a piece of network in not the same way). To do this simple default implem we were missing a getNetworkElement in Action interface and an implem of equals (using all attributs) for all Action (not only all Action used in rao). For now, without the FIXEDME, it works for all rao Action.

in the current version of the code, we put isCompatibleWith in the interface level to suggest that the behaviour should not be implementation-dependant. But it is not necessary to have a 100% generic code, I'm okay with keeping "instanceof"s, since this "isCompatibleWith" check is very RAO-centric.

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/NetworkActionImpl.java
pjeanmarie: As for other use case of Connectable, I should put back the original Branch code instead:

```
        Identifiable<?> element = network.getIdentifiable(networkElement.getId());
        if (element instanceof Branch<?> branch) {
            if (actionType == ActionType.OPEN) {
                // Line is considered closed if both terminal are connected
                return branch.getTerminal1().isConnected() && branch.getTerminal2().isConnected();
            } else {
                // Line is already considered opened if one of the terminals is disconnected
                return !branch.getTerminal1().isConnected() || !branch.getTerminal2().isConnected();
            }
        } else if (element instanceof Switch sw) {
            return sw.isOpen() == (actionType == ActionType.CLOSE);
        } else {
            throw new NotImplementedException("Topological actions are only on branches or switches for now");
        }
```

data/crac-creation/crac-creator-cim/src/main/java/com/powsybl/openrao/data/craccreation/creator/cim/craccreator/NetworkActionCreator.java
```diff
@@ -81,15 +80,19 @@ public void createNetworkActionAdder() {
                 if (psrType.equals(CimConstants.PsrType.PST.getStatus())) {
                     // PST setpoint elementary action
                     addPstSetpointElementaryAction(elementaryActionId, remedialActionRegisteredResource, networkActionAdder);
-                } else if (psrType.equals(CimConstants.PsrType.GENERATION.getStatus()) || psrType.equals(CimConstants.PsrType.LOAD.getStatus())) {
-                    // Injection elementary action
-                    addInjectionSetpointElementaryAction(elementaryActionId, remedialActionRegisteredResource, networkActionAdder);
+                } else if (psrType.equals(CimConstants.PsrType.GENERATION.getStatus())) {
+                    addGeneratorElementaryAction(elementaryActionId, remedialActionRegisteredResource, networkActionAdder);
+                    // ------ TODO : Missing check : default capacity in ohms
```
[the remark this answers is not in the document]
pjeanmarie: removed

data/crac-creation/crac-creator-cim/src/main/java/com/powsybl/openrao/data/craccreation/creator/cim/craccreator/NetworkActionCreator.java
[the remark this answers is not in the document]
pjeanmarie: fixed Connectable -> Branch. no change for side.

data/crac-creation/crac-creator-cse/src/main/java/com/powsybl/openrao/data/craccreation/creator/cse/remedialaction/TRemedialActionAdder.java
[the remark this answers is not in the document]
pjeanmarie: fixed Connectable -> Branch

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/NetworkActionImpl.java
pjeanmarie: fixed Connectable -> Branch

data/crac-creation/crac-creator-cim/src/main/java/com/powsybl/openrao/data/craccreation/creator/cim/craccreator/NetworkActionCreator.java
```diff
@@ -127,13 +128,13 @@ private void addPstSetpointElementaryAction(String elementaryActionId, RemedialA
         } else {
             throw new OpenRaoImportException(ImportStatus.INCONSISTENCY_IN_DATA, String.format("Wrong marketObjectStatusStatus: %s on elementary action %s", marketObjectStatusStatus, elementaryActionId));
         }
-        networkActionAdder.newPstSetPoint()
+        networkActionAdder.newPhaseTapChangerTapPositionAction()
             .withNetworkElement(networkElementId)
-            .withSetpoint(setpoint)
+            .withTapPosition((pstHelper.getLowTapPosition() + pstHelper.getHighTapPosition()) / 2 + setpoint)
```
annetill, line 133: We have in the API a method `getNeutralPosition()` that gives the tap where rho = 1 and alpha = 0 if exists. What is the initial need?

data/crac-creation/crac-creator-cse/src/main/java/com/powsybl/openrao/data/craccreation/creator/cse/remedialaction/TRemedialActionAdder.java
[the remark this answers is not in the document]
annetill: Thanks

data/crac-io/crac-io-json/src/main/java/com/powsybl/openrao/data/craciojson/deserializers/InjectionSetpointArrayDeserializer.java
```diff
@@ -26,37 +28,88 @@ public final class InjectionSetpointArrayDeserializer {
...
-                            adder.withNetworkElement(networkElementId, networkElementsNamesPerId.get(networkElementId));
-                        } else {
-                            adder.withNetworkElement(networkElementId);
-                        }
+                        networkElementId = jsonParser.nextTextValue();
                         break;
                     case SETPOINT:
                         jsonParser.nextToken();
-                        adder.withSetpoint(jsonParser.getIntValue());
+                        setpoint = jsonParser.getIntValue();
```
obrix, line 52: Should be a double.

data/crac-creation/crac-creator-cim/src/main/java/com/powsybl/openrao/data/craccreation/creator/cim/craccreator/NetworkActionCreator.java
annetill: Ok, you are just iso-functional with this line.

data/crac/crac-api/src/test/java/com/powsybl/openrao/data/cracapi/NetworkActionUtils.java
```diff
@@ -372,19 +236,25 @@ public static NetworkElement createNetworkElement(String networkElementId) {
...
 
-    public static PstSetpoint createPstSetpoint(NetworkElement pst, int setpoint) {
-        return new PstSetpointImplTest(pst, setpoint);
+    public static NetworkAction createPhaseTapChangerTapPositionAction(NetworkElement pst, int setpoint) {
+        return new NetworkActionImplTest(new PhaseTapChangerTapPositionActionBuilder().withId("id").withNetworkElementId(pst.getId()).withTapPosition(setpoint).withRelativeValue(false).build());
     }
 
-    public static InjectionSetpoint createInjectionSetpoint(NetworkElement networkElement, double setpoint, Unit unit) {
-        return new InjectionSetpointImplTest(networkElement, setpoint, unit);
+    public static NetworkAction createGeneratorAction(NetworkElement networkElement, double setpoint, Unit unit) {
```
obrix, line 251: Not very important because that's only used in test but base the selection of action on a Uni is not very explicit.
 It may be a good idea to split into two function, createGeneratorTargetVAction and createGeneratorActivePowerAction

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/NetworkActionImpl.java
```diff
@@ -7,55 +7,120 @@
...
             return false;
         } else {
-            elementaryActions.forEach(action -> action.apply(network));
+            elementaryActions.forEach(action -> action.toModification().apply(network, true, ReportNode.NO_OP));
             return true;
         }
     }
 
+    @Override
+    public boolean canBeApplied(Network network) {
```
obrix, line 98: @annetill To check

data/crac-creation/crac-creator-cse/src/main/java/com/powsybl/openrao/data/craccreation/creator/cse/remedialaction/TRemedialActionAdder.java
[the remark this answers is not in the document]
annetill: @pet-mit no three windings transformer in the RAO?

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/SwitchActionAdderImpl.java
```diff
@@ -0,0 +1,50 @@
...
+    public SwitchActionAdder withActionType(ActionType actionType) {
+        this.actionType = actionType;
+        return this;
+    }
+
+    protected Action buildAction() {
+        return new SwitchActionBuilder()
+            .withId(String.format("%s_%s_%s", getActionName(), networkElementId, actionType))
+            .withNetworkElementId(networkElementId)
+            .withOpen(actionType == ActionType.OPEN)
```
obrix, line 39: This will raise a nullptr exception if actionType is missing. Use patterne ActionType.OPEN.equals(actionType) to avoid that. There is several places in the code where you can use this pattern instead of "==".

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/NetworkActionImpl.java
annetill: It is iso functional with main, but questionnable.

data/crac-creation/crac-creator-cse/src/main/java/com/powsybl/openrao/data/craccreation/creator/cse/remedialaction/TRemedialActionAdder.java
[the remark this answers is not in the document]
pet-mit: > @pet-mit no three windings transformer in the RAO?

no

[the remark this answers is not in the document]
phiedw: CSE processes use UCTE networks, and I don't think it's possible to define three windings transformers in that format.

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/GeneratorActionAdderImpl.java
phiedw: Relative value changes are not currently supported in the RAO because it adds quite a lot of complexity (currently we can apply the same action multiple times and it will have the same effect as applying it once, which would not be the case with a relative change)

phiedw (changes requested):

data/crac-creation/crac-creator-cim/src/main/java/com/powsybl/openrao/data/craccreation/creator/cim/craccreator/NetworkActionCreator.java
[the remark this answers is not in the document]
pjeanmarie: About side: in Rao, no side input in topological action, it was hard coded that both side are affected (open or close), for terminal-connection-action, no side means all side, so its the same

data/crac-creation/crac-creator-fb-constraint/src/main/java/com/powsybl/openrao/data/craccreation/creator/fbconstraint/ActionReader.java
```diff
@@ -101,16 +107,25 @@ void addAction(PstRangeActionAdder pstRangeActionAdder) {
...
 
         if (!type.equals(Type.TOPO)) {
             throw new OpenRaoException(String.format("This method is only applicable for Action of type %s", TOPO));
         }
 
-        networkActionAdder.newTopologicalAction()
-            .withNetworkElement(networkElementId)
-            .withActionType(topologicalActionType)
-            .add();
+        if (networkElement.getType() == IdentifiableType.SWITCH) {
```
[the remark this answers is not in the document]
pjeanmarie: similar adder but not the same type, not the same type of adder because not the same type of core action to create

data/crac-io/crac-io-json/src/main/java/com/powsybl/openrao/data/craciojson/deserializers/InjectionSetpointArrayDeserializer.java
pjeanmarie: For iso functionnality, it was read as int (see line above)...

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/TerminalsConnectionActionAdderImpl.java
```diff
@@ -0,0 +1,49 @@
...
+    public TerminalsConnectionActionAdder withActionType(ActionType actionType) {
+        this.actionType = actionType;
+        return this;
+    }
+
+    protected Action buildAction() {
+        return new TerminalsConnectionActionBuilder()
+            .withId(String.format("%s_%s_%s", getActionName(), networkElementId, actionType))
+            .withNetworkElementId(networkElementId)
+            .withOpen(actionType == ActionType.OPEN)
```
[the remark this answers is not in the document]
pjeanmarie: Is it possible that an exception is a better behavior than a silente missing action-type? there is also a protection against action type null (code below) but even if this protection is missing, I still have my question above

```
    protected void assertSpecificAttributes() {
        assertAttributeNotNull(actionType, getActionName(), "actionType", "withActionType()");
    }
```

pjeanmarie:
NB. new class MaximumNumberOfElementaryActionsFilter

data/crac/crac-impl/src/test/java/com/powsybl/openrao/data/cracimpl/utils/ExhaustiveCracCreation.java
```diff
@@ -202,10 +202,28 @@ public static Crac create(CracFactory cracFactory) {
...
             .withName("injectionSetpointRaName")
             .withOperator("RTE")
-            .newInjectionSetPoint().withSetpoint(260).withNetworkElement("injection").withUnit(Unit.SECTION_COUNT).add()
+            .newGeneratorAction().withActivePowerValue(260.0).withNetworkElement("injection").add()
             .newOnConstraintUsageRule().withCnec("cnec3autoId").withInstant(AUTO_INSTANT_ID).withUsageMethod(UsageMethod.FORCED).add()
             .add();
 
+        // network action with multiple type elementary actions
+        crac.newNetworkAction().withId("injectionSetpointRa2Id")
+            .withName("injectionSetpointRa2Id")
```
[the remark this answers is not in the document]
pjeanmarie: done

data/crac/crac-impl/src/test/java/com/powsybl/openrao/data/cracimpl/utils/ExhaustiveCracCreation.java
```diff
@@ -202,10 +202,28 @@ public static Crac create(CracFactory cracFactory) {
...
+            .withOperator("RTE")
+            .newLoadAction().withActivePowerValue(260.0).withNetworkElement("LD1").add()
+            .newDanglingLineAction().withActivePowerValue(-120.0).withNetworkElement("DL1").add()
+            .newSwitchAction().withActionType(ActionType.OPEN).withNetworkElement("BR1").add()
+            .newOnFlowConstraintInCountryUsageRule().withInstant(CURATIVE_INSTANT_ID).withContingency("contingency2Id").withCountry(Country.FR).withUsageMethod(UsageMethod.AVAILABLE).add()
+            .add();
+
+        // network action with shunt compensator elementary action
+        crac.newNetworkAction().withId("injectionSetpointRa3Id")
+            .withName("injectionSetpointRa3Id")
```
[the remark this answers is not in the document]
pjeanmarie: done

docs/input-data/crac/json.md
```diff
@@ -834,67 +834,102 @@ OpenRAO handles three different types of usage methods sorted by priority:
...
-          .withNetworkElement("shunt-compensator-id")
-          .withUnit(Unit.SECTION_COUNT)
-          .add()
-        .newOnInstantUsageRule().withUsageMethod(UsageMethod.AVAILABLE).withInstant(PREVENTIVE_INSTANT).add()
-        .add();
+// shunt compensator position action
+crac.newNetworkAction()
+    .withId("injection-setpoint-shunt-compensator-id")
+    .withOperator("operator")
+    .newInjectionSetPoint()
```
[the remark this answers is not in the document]
pjeanmarie: done

docs/input-data/crac/json.md
```diff
@@ -834,67 +834,102 @@ OpenRAO handles three different types of usage methods sorted by priority:
...
-        .newInjectionSetPoint()
-          .withSetpoint(3)
-          .withNetworkElement("shunt-compensator-id")
-          .withUnit(Unit.SECTION_COUNT)
-          .add()
-        .newOnInstantUsageRule().withUsageMethod(UsageMethod.AVAILABLE).withInstant(PREVENTIVE_INSTANT).add()
-        .add();
+// shunt compensator position action
+crac.newNetworkAction()
+    .withId("injection-setpoint-shunt-compensator-id")
```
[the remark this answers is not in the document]
pjeanmarie: done

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/SwitchPairImpl.java
```diff
@@ -7,44 +7,42 @@
...
+                    .withNetworkElementId(switchToOpen.getId())
+                    .withOpen(true)
+                .build()
+                .toModification(),
+            new SwitchActionBuilder()
+                .withId("idClose")
+                .withNetworkElementId(switchToClose.getId())
+                .withOpen(false)
+            .build()
+            .toModification()
```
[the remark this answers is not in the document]
pjeanmarie: done

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/NetworkActionImpl.java
```diff
@@ -7,55 +7,120 @@
...
+        // TODO: To implement on powsybl-core Action
+        return elementaryActions.stream().allMatch(elementaryAction -> {
+            if (elementaryAction instanceof ShuntCompensatorPositionAction shuntCompensatorPositionAction) {
+                ShuntCompensator shuntCompensator = network.getShuntCompensator(shuntCompensatorPositionAction.getShuntCompensatorId());
+                return shuntCompensatorPositionAction.getSectionCount() <= shuntCompensator.getMaximumSectionCount();
+            } else if (elementaryAction instanceof GeneratorAction || elementaryAction instanceof LoadAction || elementaryAction instanceof DanglingLineAction) {
+                return true;
+            } else if (elementaryAction instanceof PhaseTapChangerTapPositionAction) {
+                // TODO : setpoint out of range ?
```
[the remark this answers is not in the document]
pjeanmarie: If we can change a bit the behavior of canBeApplied (compared to main) and based on my understanding of your needs, I direclty used the check system of core, you can see the code and the impact on test (real impact only on the pst out or range check):
https://github.com/powsybl/powsybl-open-rao/pull/920/commits/2bae784302fb68c223adb1e92dcc2c47c0cbe2bf

data/crac-io/crac-io-json/src/main/java/com/powsybl/openrao/data/craciojson/deserializers/InjectionSetpointArrayDeserializer.java
obrix: Yes I saw that but isn't it a bug ? We are setting a injection setpoint, all double value in json will be truncated ?

data/crac/crac-api/src/test/java/com/powsybl/openrao/data/cracapi/NetworkActionUtils.java
pjeanmarie: done

data/crac-io/crac-io-json/src/main/java/com/powsybl/openrao/data/craciojson/deserializers/InjectionSetpointArrayDeserializer.java
pjeanmarie: It will be truncated that's why I did not change that, I wanted to avoid functional change as much as possible in case it breaks the tests (especially the ones not run by the CI). But I added the fix for that, we will see. (I can also do a separate PR for this bug if we exêct it to break the tests).

phiedw: It does make more sense with a double indeed. If a test breaks because of it, then we will fix the test.

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/NetworkActionImpl.java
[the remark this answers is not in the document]
phiedw: The change suggested seems dangerous. You don't really expect canBeApplied to actually apply the action, and if I'm not mistaken with this change it would.

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/NetworkActionImpl.java
```diff
@@ -7,55 +7,114 @@
...
 
+    @Override
+    public boolean canBeApplied(Network network) {
+        return elementaryActions.stream().allMatch(elementaryAction -> {
+            if (elementaryAction instanceof SwitchPair switchPair) {
+                // It is only applicable if, initially, one switch was closed and the other was open.
+                return network.getSwitch(switchPair.getSwitchToOpen().getId()).isOpen() != network.getSwitch(switchPair.getSwitchToClose().getId()).isOpen();
+            } else {
+                try {
+                    elementaryAction.toModification().apply(network, true, ReportNode.NO_OP);
```
phiedw, line 106: applying the action in a method called canBeApplied may be dangerous in the future

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/TerminalsConnectionActionAdderImpl.java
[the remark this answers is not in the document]
phiedw: Right I think indeed it might actually be better to have an exception there instead of some kind of default behaviour in case the field is missing.

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/NetworkActionImpl.java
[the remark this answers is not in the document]
pjeanmarie: Yes indeed, I revert that.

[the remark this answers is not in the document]
pjeanmarie: @phiedw 
Can the transformer for the PhaseTapChangerTapPositionAction be a three WT or only two WT (here for Rao)?

pet-mit (approved):

PR #951, opened 2024-04-02
Use maven installation of or-tools and add or-tools load lib in code

pjeanmarie:

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/.github/workflows/run_cucumber_tests.yml
+++ b/.github/workflows/run_cucumber_tests.yml
@@ -13,18 +13,10 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 17
-      - name: Download OR-Tools
-        run: |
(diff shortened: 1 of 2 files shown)


obrix:
@pet-mit After some investigations it seems or tools method Loader.loadNativeLibraries does the same thing as open rao, a System.loadLibrary("jniortools"), but it also has a fallback that try to find the library into ortools resources and load it from there. That is why it works without having to install it or provide a PATH.

See ortools loader code here : https://github.com/google/or-tools/blame/stable/ortools/java/com/google/ortools/Loader.java

I think what pauline did here is the best way to load the dependency.

pet-mit:
looks OK at first glance, gonna do some tests and get back to you

pet-mit (approved):

PR #955, opened 2024-04-03
Don't skip cucumber tests

pet-mit:
Now that OR-Tools installation is not necessary, we can run cucumber tests in the Maven workflow


pet-mit:
unnecessary

PR #971, opened 2024-04-15
Fix memory leak due to leaves sometimes keeping reference to the network

phiedw:

No

Bug fix

The leaves will sometimes keep a reference to the network (for instance if the root leaf is secure before optim). These are then stored in the final result and can cause memory issues

We now make sure to remove network references from the leaves

- [ ] Yes
- [x] No


ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/searchtree/algorithms/SearchTree.java
```diff
@@ -370,7 +374,6 @@ private void optimizeLeaf(Leaf leaf) {
         } else {
             TECHNICAL_LOGS.info("No range actions to optimize");
         }
-        leaf.finalizeOptimization();
```
pet-mit, line 373: why not keep this here to minimize the memory footprint of the search-tree, and only add it for lines 117 & 123?

phiedw: The networks are kept in memory anyways during the search tree by the network pool.
It's only important to make sure we remove the reference to the network when returning the leaf that will be stored in the results.

phiedw: (and as an example of the complexity, we'd also have to call it at line 345 if we don't call it at the top level, because of leaves that reach the stop criterion before optimization, with just the topo action applied)

pet-mit (approved):

PR #972, opened 2024-04-17
[CSA] [Import] Reorder usage rules definition

bqth29:

Usage rules are created by decreasing order or restriction:
1. first, the OnConstraint usage rules are created
2. then, if the RA is linked to no AE, OnContingencyState usage rules are created
3. finally, if the RA is linked to no AE and no CO, the OnInstant usage rules are created

> If some AEs have the isCombinableWithRemedialAction set to true, OnConstraint usage rules can be created too. The reason for this is to avoid undesired behaviors of a TSO defining an AE combinable with remedial actions that would make the RA of another TSO less available than what is was designed for (ex: onFlowConstraint instead of onInstant)
>
> If a RA is linked to contingencies and assessed elements (implicitly or explicitly), OnConstraint usage rules are created only for the CNECs that are defined for these very contingencies


data/crac-creation/crac-creator-csa-profiles/src/main/java/com/powsybl/openrao/data/craccreation/creator/csaprofile/craccreator/CsaProfileCracUtils.java
```diff
@@ -30,10 +30,10 @@ private CsaProfileCracUtils() {
 
     }
 
-    public static Map<String, Set<PropertyBag>> getMappedPropertyBagsSet(PropertyBags propertyBags, String property) {
+    public static Map<String, Set<PropertyBag>> groupPropertyBagsBy(PropertyBags propertyBags, String groupBy) {
         Map<String, Set<PropertyBag>> mappedPropertyBags = new HashMap<>();
         for (PropertyBag propertyBag : propertyBags) {
-            String propValue = propertyBag.getId(property);
+            String propValue = propertyBag.getId(groupBy);
```
benrejebmoh, line 36:
```suggestion
            String propValue = propertyBag.getId(groupingProperty);
```

Issue #890, opened 2024-02-20
Update GLSK handling for AngleMonitoring


pet-mit:
### Describe the current behavior

In the angle monitoring module, re-dispatching is applied by stopping generators identified in the remedial actions linked to the angle constraints, then re-dispatching the lost generation on the other generators in the GLSK.
In order to excluded the stopped generators from the GLSK before asking powsybl to re-dispatch the lost generation, specific code is written.

### Describe the expected behavior

Use powsybl's new ability to exclude generators from GLSK

### Describe the motivation

Cleaner code

Godelaine:
merged : https://github.com/powsybl/powsybl-open-rao/pull/1092

PR #993, opened 2024-04-30
Angle Monitoring with Scalable

pet-mit:

Feature

Angle Monitoring uses CIM GLSK document to re-dipatch lost generation after stopping generators. 

Now it should be able to use Scalable which is more generic.
Unfortunately, Scalable cannot be filtered like CimGLSKDocument: we need to exclude from the scaling generators that were stopped by the angle monitoring module when redispatching their production.
This requires the next release of powsybl that will add in scaling parameters a set of injections to ignore.
And also closes issue here: https://github.com/powsybl/powsybl-open-rao/issues/890 

- [x] Yes
- [ ] No

AngleMonitoring API using CimGlskDocument is deprecated. 
The users should convert the CimGlskDocument to a Scalable, using the Instant, and use the new API.
~~~java
new AngleMonitoring(crac, network, raoResult, cimGlskDocument.getZonalScalable(inputNetwork, Instant.from(glskOffsetDateTime))
~~~


pet-mit:
now possible thanks to https://github.com/powsybl/powsybl-core/pull/3010

however, might be done as part of this refactoring: https://github.com/powsybl/powsybl-open-rao/issues/899

PR #1078, opened 2024-07-05
Fix OFCC triggered after any outage

bqth29:

A bug was reported from the SWE CC process where an OnFlowConstraintInCountry auto remedial action with no contingency defined was systematically triggered after all outages whereas only some of them were actually creating an overload in the given country.

The CNECs retrieved in the automaton simulator were not filtered on the state so this was corrected. Two Cucumber tests were added (one for ARA one for CRA) to ensure the problem is fixed.


ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/castor/algorithm/AutomatonSimulator.java
```diff
@@ -249,7 +249,7 @@ public Set<NetworkAction> getActivatedNetworkActions() {
     TopoAutomatonSimulationResult simulateTopologicalAutomatons(State automatonState, Network network, PrePerimeterSensitivityAnalysis preAutoPstOptimizationSensitivityAnalysis) {
         // -- Apply network actions
         // -- First get forced network actions
-        Set<FlowCnec> flowCnecs = crac.getFlowCnecs();
+        Set<FlowCnec> flowCnecs = crac.getFlowCnecs(automatonState);
```
phiedw, line 252: should we also check the curative cnecs on the same contingency?

phiedw: (to check with Sebastien)

murgeyseb: It is a good question... I would say no, but still have to check

pet-mit (approved):

PR #1079, opened 2024-07-05
enable back preventive factors after sensi ends in 2P

MartinBelthle:

No

Bug fix

In 2nd+ MIP iterations in 2P optimization, we had 0 flow on all preventive cnecs.
This caused too-optimistic decisions (0 flow means lot of margin) except when considering loop-flows where we'd have huge loop-flows penalties (0 flow but 1000MW of commercial flow means 1000MW of loop-flows) .

We did so because we disabled the preventive factors to speed-up sensi calculations on curative states without re-enabling it after all calculations were over.

We re-enable the factors after the sensi calculations and therefore do not have this problem.

- [ ] Yes
- [x] No

**Other information**:
Cucumber test are running with this branch to see how many real cases were impacted.


phiedw (commented):
Maybe we should add a small non regression test to be safe

phiedw:
> Maybe we should add a small non regression test to be safe

It's actually quite difficult to create such a test case on a 12node network, since it requires the MIP giving different results throughout the iterations, so I guess we should just keep the case that revealed the problem in our sensitive test data.

sensitivity-analysis/src/main/java/com/powsybl/openrao/sensitivityanalysis/SystematicSensitivityAdapter.java
```diff
@@ -130,6 +130,9 @@ static SystematicSensitivityResult runSensitivity(Network network,
             network.getVariantManager().removeVariant(variantForState);
         }
 
+        // enable preventive factors for next iterations
+        cnecSensitivityProvider.enableFactorsForBaseCaseSituation();
```
pet-mit, line 134: if i comment this line, the unit test that you added does not fail

pet-mit (changes requested):

PR #1086, opened 2024-07-09
More robust MIP + OR-Tools bump

pet-mit:

Bug fix

MIP resolution can have unexpected behaviour (eg give out a suboptimal solution) when using too large numbers that do not exceed its own definition of infinity, or too small numbers that are not zero (because of scaling algorithms).

Now "infinity" value is defined solver-wise, and used where suitable. For instance, for XPRESS we should use 1e20 instead of 1e10.
Plus, maxTso & maxRa constraint is skipped for cases where the maxTso/maxRa value is too large (and useless), in order to avoid badly conditionned MIPs.
Also, using infinity showed there was a bug in or-tools v9.9, so we bumped up to v9.10.

- [ ] Yes
- [x] No


phiedw (approved):
great work!

PR #1087, opened 2024-07-09
Add new method withNetworkElement for InjectionRangeActionAdder

wangjer:

Feature

Adding a new network element to an "InjectionRangeAction" requires to use "withNetworkElementAndKey" method.

We can now use withNetworkElement method instead, so that "key" argument is set to 1.0 by default

- [ ] Yes
- [ ] No

**Other information**:


data/crac/crac-impl/src/test/java/com/powsybl/openrao/data/cracimpl/InjectionRangeActionAdderImplTest.java
```diff
@@ -125,6 +125,37 @@ void testAddWithoutGroupId() {
...
     }
 
+    @Test
+    void testAddDefaultKey() {
+        InjectionRangeAction injectionRangeAction = crac.newInjectionRangeAction()
+            .withId("id1")
+            .withOperator("BE")
+            .withGroupId("groupId1")
+            .withNetworkElement(injectionId1)
+            .withNetworkElement(injectionId2, injectionName2)
```
pet-mit, line 135: I wonder if we want this to add both elements with key = 1, or only the last added element (i.e. injectionId2), with key = 1. what do you think @phiedw ?

pet-mit: @wangjer can you please update the branch following these comments?

phiedw (dismissed):

bqth29 (approved):

PR #1093, opened 2024-07-15
MIP bug fixes

pet-mit:
Fix a bug introduced in #1090, by correcting the way integer variations of PSTs are taken into account between 2 MIP iterations


ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/linearoptimisation/algorithms/fillers/DiscretePstTapFiller.java
```diff
@@ -93,17 +96,16 @@ private void buildPstTapVariablesAndConstraints(LinearProblem linearProblem, Pst
...
-            relativeTapConstraintUpward.setCoefficient(pstTapUpwardVariationVariable, 1);
-            relativeTapConstraintUpward.setCoefficient(preventivePstTapUpwardVariationVariable, -1);
-
-            OpenRaoMPConstraint relativeTapConstraintDownard = linearProblem.addPstRelativeTapConstraint(0, minRelativeTap, pstRangeAction, state, LinearProblem.VariationDirectionExtension.DOWNWARD);
             OpenRaoMPVariable preventivePstTapDownwardVariationVariable = linearProblem.getPstTapVariationVariable((PstRangeAction) preventiveRangeAction, optimizationPerimeter.getMainOptimizationState(), LinearProblem.VariationDirectionExtension.DOWNWARD);
-            relativeTapConstraintDownard.setCoefficient(pstTapDownwardVariationVariable, 1);
-            relativeTapConstraintDownard.setCoefficient(preventivePstTapDownwardVariationVariable, -1);
+
+            OpenRaoMPConstraint relativeTapConstraint = linearProblem.addPstRelativeTapConstraint(minRelativeTap, maxRelativeTap, pstRangeAction, state);
```
phiedw, line 104: the bounds of this constraint need to take into consideration the "initial" tap of the psts in the fill method as well

phiedw (changes requested):

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/linearoptimisation/algorithms/fillers/DiscretePstTapFiller.java
pet-mit: in the fill method (1st iteration), both taps are equal (= to the tap in the input network), aren't they? so they cancel each other out.
however I think we should take them into account in the update method

phiedw (approved):

MartinBelthle (approved):

PR #1124, opened 2024-09-06
[CSA Importer] Read CNECs borders

bqth29:

CNECs were always created with a null border.

According to what was decided for the CSA SWE process, the border to which an AssessedElement is linked can be specified though the field `AssessedElement.OverlappingZone`. 

Only REE needs to specify this information since REN and RTE both only have one border so the AssessedElement's operator is enough to know said border.

A new CSA CRAC creation parameters field called `"borders"` was added to indicate how to map an AssessedElement to a specific border.

```json
"borders": [
  {
    "name": "ES-FR",
    "eic": "10YDOM--ES-FR--D",
    "default-for-tso": "RTE"
  },
  {
    "name": "ES-PT",
    "eic": "10YDOM--ES-PT--T",
    "default-for-tso": "REN"
  }
]
```
- `name` is the name of the border that will appear in the `border` field of the CNECs in the CRAC
- `eic` is the EI Code of the border
- `default-for-tso` indicates for which operator (TSO) this border is the default border (used for TSOs with only one border)

When an AssessedElement is being parsed by the importer the OverlappingZone field is read:
- if it is not null then its border is the border with the same EI Code (or null if no such border were defined in the CRAC creation parameters)
- if it is null, then its border is the default border for the AssessedElement's operator (or null is there is no default border for this operator)

- [ ] Yes
- [x] No


data/crac-io/crac-io-csa-profiles/src/main/java/com/powsybl/openrao/data/cracio/csaprofiles/parameters/CsaCracCreationParameters.java
```diff
@@ -9,16 +9,25 @@
...
 
 /**
  * @author Mohamed Ben-rejeb {@literal <mohamed.ben-rejeb at rte-france.com>}
  */
 public class CsaCracCreationParameters extends AbstractExtension<CracCreationParameters> {
+    private static final String REE = "REE";
+    private static final String REN = "REN";
+    private static final String RTE = "RTE";
+    private static final String ES_FR = "ES-FR";
+    private static final String ES_PT = "ES-PT";
```
jipea, line 25: I think some of these constants strings are already in craccreator.constants package. If there aren't, I think you should add it.

jipea (changes requested):

jipea (approved):

PR #1130, opened 2024-09-11
[CSA] No longer read AE.isCombinableWithRemedialAction attribute

bqth29:

The field isCombinableWithRemedialAction was making the CRAC heavier than needed by adding OnConstraint usage rules even though the behavior is already covered by other usage rules.

The attribute is no longer read which avoid creating several useless or undesired OnConstraint usage rules for remedial actions.


data/crac/crac-io/crac-io-csa-profiles/src/main/java/com/powsybl/openrao/data/cracio/csaprofiles/craccreator/cnec/FlowCnecCreator.java
```diff
@@ -47,6 +47,13 @@ public FlowCnecCreator(Crac crac, Network network, AssessedElement nativeAssesse
         checkCnecDefinitionMode();
     }
 
+    private static Set<Country> getBranchLocation(Branch<?> branch) {
```
jipea, line 50: I think this method is useless now, because there's no more geographical filter

jipea (changes requested):

jipea (dismissed):

benrejebmoh (approved):

PR #1133, opened 2024-09-12
Add cost as new objective function for linear problem 

wangjer:

- [ ] Yes
- [ ] No

**Other information**:


wangjer:
The way the new objective function is implemented is flawed.
Right now, the min margin variable $MM$ is supposed to be equal to 0 if all cnecs are secure.
However, in practice, the linear problem finds a solution that is very slightly unsecure, putting a negative Min Margin, despite the big penalty.
That is why the cucumber test "Preventive RA with minimal cost" failed.

bqth29:
overridden by #1243

PR #1140, opened 2024-09-18
Rao parameters use unit from is dc

pjeanmarie:

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/data/result-exporter/core-cne-exporter/src/test/java/com/powsybl/openrao/data/corecneexporter/CoreCneCnecsCreatorTest.java
+++ b/data/result-exporter/core-cne-exporter/src/test/java/com/powsybl/openrao/data/corecneexporter/CoreCneCnecsCreatorTest.java
@@ -171,7 +171,7 @@ class CoreCneCnecsCreatorTest {
 
         mockCnecResult(cnec2, 800, -200, -999999999, .2);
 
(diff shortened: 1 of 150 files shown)


tests/src/test/resources/com/powsybl/openrao/tests/features/epic10_relative_margin/US10_1.feature
```diff
@@ -5,21 +5,21 @@
 
 Feature: US 10.1: Linear RAO with relative margin
 
-  @fast @rao @mock @ac @preventive-only @relative
+  @fast @rao @mock @dc @preventive-only @relative
```
bqth29, line 8: Is there a reason you moved from AC to DC? I am affraid this will change the LF results which could alter the test's results

bqth29: The same question goes for all the other tests, I would rather convert by hand the objective value from MW to A instead of changing the LF computation mode

what do you think @phiedw?

pjeanmarie: This dev will be replaced by a separate unit parameter in rao parameter, then no change in tests behavior is expected.

pjeanmarie:
replaced by: https://github.com/powsybl/powsybl-open-rao/pull/1197

phiedw:
replaced by: https://github.com/powsybl/powsybl-open-rao/pull/1197

Issue #1136, opened 2024-09-16
RaoResult writer API


pet-mit:
### Describe the current behavior

RaoResult can be exported to json, core cne, or swe cne. Three APIs exist for those purposes.

### Describe the expected behavior

Unify the API by creating one method : raoResult.write

### Describe the motivation

cleaner API, ease to add new exporters, ease to develop python api

### Extra Information

- The method has the following arguments : CracCreationContext, Properties (that contains all specific information needed for specific exporters)
- The CNE exporters shall no longer need the Network : some preparatory work might be needed (for instance, adding information to the CracCreationContext)
- If it's useful, we can add a second method that uses a Crac instead of CracCreationContext
- If it's a lot of work, try to split it up in multiple Pull Requests

PR #1157, opened 2024-10-07
CNE export no longer uses network

bqth29:

To prepare the common RaoResult exporter using `RaoResult.write` and as presented in #1136, the CNE exporters no longer require the network as an input and the relevant information was set in the CRAC Creation Context instead.

Need of network to export RaoResult as CNE.

No longer need of network to export RaoResult as CNE.

⚠️ **Breaking change**: the `exportCne` method no longer requires the `Network network` parameter.


data/result-exporter/swe-cne-exporter/src/main/java/com/powsybl/openrao/data/swecneexporter/SweMonitoredSeriesCreator.java
```diff
@@ -164,7 +164,7 @@ private MonitoredSeries generateMonitoredSeriesFromScratch(MonitoredSeriesCreati
     }
 
     void setInOutAggregateNodes(String networkElementId, String monitoredSeriesId, MonitoredRegisteredResource registeredResource) {
-        Branch<?> branch = sweCneHelper.getNetwork().getBranch(networkElementId);
+        Branch<?> branch = cracCreationContext.getNetworkBranches().get(networkElementId);
```
phiedw, line 167: we lose the aliases by doing it like this so we'll have to make sure it all works correctly still

bqth29: all tests, including business tests, passed

phiedw (approved):

PR #1162, opened 2024-10-14
[CSA] Use GridStateAlterations mRIDs

### LINKED ISSUE #1147: [CSA] Use GridStateAlteration mRID as elementary network action id

_No response_

Now that we use PowSyBl's RemedialAction API for which elementary actions can have their own identifier, it would make sense refactoring the CSA importer such that the elementary action are imported with the mRID of the GridStateAlteration (and the parent remedial actions keeps the GridStateAlterationRemedialAction's mRID)

### Describe the motivation

Would be useful if we plan to implement a RAS profile exporter since the mRIDs of elementary actions are required.

### Extra Information

Will add lot of text to JSON CRAC files and make them heavier. See if this cannot be avoided with a clever use of CRAC creation context.

bqth29:

Fixes #1147 

Possibility to add elementary network actions with an identifier using `withId` in adder.
If no id provided, a default id is set using the pattern `ActionClass_networkElement_setPoint`

**Other information**:
JSON CRAC updated from version 2.5 to 2.6; now includes optional `actionId` field in elementary action objects.


bqth29:
One question: if the id of an elementary action is the default one, should we export it in the JSON CRAC or not?

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/cracimpl/AbstractSingleNetworkElementActionAdderImpl.java
```diff
@@ -50,4 +56,8 @@ public NetworkActionAdder add() {
     protected abstract void assertSpecificAttributes();
 
     protected abstract String getActionName();
+
+    protected String createActionName(Object specificAttribute) {
+        return id == null ? String.format("%s_%s_%s", getActionName(), networkElementId, specificAttribute) : id;
```
phiedw, line 61: rename getActionName to getActionTypeName to avoid confusions between these two methods.

phiedw (changes requested):

bqth29:
too many conflicts, start over when less PRs are being worked on at the same time

PR #1182, opened 2024-10-23
Add remedial actions' activation cost and range actions' variation costs in API

### LINKED ISSUE #1053: Add the concept of cost in all remedial actions

Currently the only concept of cost we have on remedial actions is to slightly penalize the use of range actions in the MIP (same cost for all psts per angle and same cost for all hvdcs per MW) which are given in the parameters, to avoid using too many psts/hvdc actions for very littly gain on the minimum margin.

It would be good to have an actual cost specific to each remedial action (not per type), which would be required for instance if we want to properly consider costly remedial actions.

### Describe the motivation

Feature will be used for CORE CSA, and could be used to replace dichotomy

### Extra Information

_No response_

bqth29:

Fixes #1053 

- [ ] Yes
- [ ] No

**Other information**:
JSON CRAC version upgraded to 2.6


bqth29:
Would it be relevant to add a currency logic too?

bqth29:
We should keep in mind that the activation cost of network actions may depend on the initial se-tpoint / pre-perimeter set-point since forcing a generator to 0 from an initial set-point of 1000 MW should be more expensive than from only 50 MW.

bqth29:
PR included in #1186

PR #1186, opened 2024-10-30
Refactor objective function evaluators API and add cost-wise Search-Tree

### LINKED ISSUE #1053: Add the concept of cost in all remedial actions

bqth29:

Fixes #1053

**Refactoring of objective function evaluators and results**

evaluators

Previously, `ObjectiveFunctionResultImpl` was a non final object that had its ObjectiveFunction as an attribute so it could be recomputed when contingencies need to be filtered out (in case curative perimeters fail for instance). This was quite ambiguous for a result object that should only hold immutable data.

In the new version, the results are decomposed based on the states and can be filtered in case contingencies need to be excluded.

results

Finally, the objective function builder was becoming useless with this refactoring and was thus replaced by creators as follows:

creator

- [ ] Yes
- [ ] No

**Other information**:


data/crac/crac-io/crac-io-json/src/main/java/com/powsybl/openrao/data/cracio/json/deserializers/PstRangeActionArrayDeserializer.java
```diff
@@ -156,4 +165,18 @@ private static void deserializeOlderOnConstraintUsageRules(JsonParser jsonParser
...
     }
+
+    private static void deserializeVariationCosts(PstRangeActionAdder pstRangeActionAdder, JsonParser jsonParser) throws IOException {
+        while (jsonParser.nextToken() != JsonToken.END_OBJECT) {
+            if (UP.equals(jsonParser.getCurrentName())) {
+                jsonParser.nextToken();
+                pstRangeActionAdder.withVariationCost(jsonParser.getDoubleValue(), RangeAction.VariationDirection.UP);
+            } else if (DOWN.equals(jsonParser.getCurrentName())) {
+                jsonParser.nextToken();
+                pstRangeActionAdder.withVariationCost(jsonParser.getDoubleValue(), RangeAction.VariationDirection.DOWN);
```
pet-mit, line 176: instead of using if/else, create a "deserialize" method (in the Constants class) that consumes a String and returns a VariationDirection

docs/input-data/crac/json.md
```diff
@@ -1187,23 +1190,27 @@ Note that the [PstHelper utility class](https://github.com/powsybl/powsybl-open-
...
-⚪ **operator**  
-🔴 **network element**: id is mandatory, name is optional  
-⚪ **groupId**: if you want to align this range action with others, set the same groupId for all. You can use any
+- 🔴⭐ **identifier**  
+- ⚪ **name**  
+- ⚪ **operator**
+- ⚪ **activationCost**
+- ⚪ **variationCosts**
+- - ⚪ **up**
+- - ⚪ **down**
```
pet-mit, line 1199: fill this out

docs/input-data/crac/json.md
```diff
@@ -1187,23 +1190,27 @@ Note that the [PstHelper utility class](https://github.com/powsybl/powsybl-open-
...
-⚪ **onState usage rules**: list of 0 to N OnContingencyState usage rules (see paragraph on [usage rules](#remedial-actions-and-usages-rules))  
-⚪ **onFlowConstraintInCountry usage rules**: list of 0 to N OnFlowConstraintInCountry usage rules (see previous paragraph on [usage rules](#remedial-actions-and-usages-rules))  
-⚪ **onConstraint usage rules**: list of 0 to N OnConstraint usage rules (see previous paragraph on [usage rules](#remedial-actions-and-usages-rules))  
+- 🔵 **speed**: mandatory if it is an automaton  
+- 🔴 **initial tap**  
+- 🔴 **tap to angle conversion map**  
+- 🔴 **tap ranges**: list of 0 to N TapRange  
+- - 🔴 **range type**  
+- - 🔵 **min tap**  
+- - 🔵 **max tap**: at least one value must be defined
```
pet-mit, line 1209: doesn't look good
image

ra-optimisation/rao-api/src/main/java/com/powsybl/openrao/raoapi/parameters/ObjectiveFunctionParameters.java
pet-mit, line 1: parameters are being refactored. Maybe wait for #1101 to be merged before managing that part?

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/FunctionalCostEvaluator.java
pet-mit, line 1: you don't need to separate FunctionalCostEvaluator from VirtualCostEvaluator. This creates extra complexity with no added value.

pet-mit (changes requested):
I feel that you haven't finished this, let me know when you do

ra-optimisation/rao-api/src/main/java/com/powsybl/openrao/raoapi/parameters/ObjectiveFunctionParameters.java
bqth29: I know, I added the parameters for development purposes only for the moment

bqth29: however I don't know if we want to wait for #1101 to be merged before this PR, if so how should we proceed?

docs/input-data/crac/json.md
bqth29: nevermind this can be fixed by adding two white spaces at the end of each line

bqth29:
> I feel that you haven't finished this, let me know when you do

This PR only focuses on costly network actions, range actions will come in another PR later
I may have unintentionally forgotten features but I believe this PR is ready

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/castor/algorithm/CastorFullOptimization.java
```diff
@@ -80,13 +81,14 @@ public CompletableFuture<RaoResult> run() {
             toolProvider);
 
         PrePerimeterResult initialOutput;
-        initialOutput = prePerimeterSensitivityAnalysis.runInitialSensitivityAnalysis(network, crac);
+        initialOutput = prePerimeterSensitivityAnalysis.runInitialSensitivityAnalysis(network, crac, Set.of());
```
pet-mit, line 84: instead of passing an empty set, create another method without the last argument

pet-mit (dismissed):

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/result/impl/PreventiveAndCurativesRaoResultImpl.java
```diff
@@ -359,6 +367,17 @@ private double getHighestFunctionalForInstant(Instant instant) {
         return highestFunctionalCost;
     }
 
+    private double getTotalFunctionalCostForInstant(Instant instant) {
+        double preventiveCost = secondPreventivePerimeterResult.getFunctionalCost();
+        return preventiveCost
+            + postContingencyResults.entrySet().stream()
```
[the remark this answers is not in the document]
bqth29: I am dealing with 2P in #1194 because there is no way to optimize linear range actions cos-wise with this current PR
but the cost is based on activated range actions so I do not think that they will be counted twice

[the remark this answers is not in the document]
bqth29: it could be worth to add a unit test with this situation in #1194 though

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/EvaluatorsUtils.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * @author Thomas Bouquet {@literal <thomas.bouquet at rte-france.com>}
+ */
+public final class EvaluatorsUtils {
+
+    private EvaluatorsUtils() {
+    }
+
+    public static List<FlowCnec> getCostlyElements(Set<FlowCnec> flowCnecs, MarginEvaluator marginEvaluator, Unit unit, FlowResult flowResult, Set<String> contingenciesToExclude) {
```
phiedw, line 29: rename this method so we know we're getting the most constrained cnecs

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/ObjectiveFunction.java
```diff
@@ -45,22 +46,22 @@ public Set<FlowCnec> getFlowCnecs() {
     }
 
     public Pair<Double, List<FlowCnec>> getFunctionalCostAndLimitingElements(FlowResult flowResult) {
```
phiedw, line 48: why can this one not be called with a flow result and a remedial action activation result?

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/ObjectiveFunction.java
```diff
@@ -107,11 +114,18 @@ public ObjectiveFunction build(Set<FlowCnec> flowCnecs,
...
+                        new MarginEvaluatorWithMarginDecreaseUnoptimizedCnecs(marginEvaluator, operatorsNotToOptimizeInCurative, prePerimeterFlowResult)));
+                }
             } else {
-                this.withFunctionalCostEvaluator(new MinMarginEvaluator(flowCnecs, raoParameters.getObjectiveFunctionParameters().getType().getUnit(), marginEvaluator));
+                if (raoParameters.getObjectiveFunctionParameters().getType().costOptimization()) {
+                    addEvaluatorsForCostlyOptimization(flowCnecs, raoParameters, optimizedStates, marginEvaluator);
+                } else {
+                    this.withFunctionalCostEvaluator(new MinMarginEvaluator(flowCnecs, raoParameters.getObjectiveFunctionParameters().getType().getUnit(), marginEvaluator));
+                }
             }
```
phiedw, line 129: it would be cleaner this way I think
if (not optimized...) {
   marginEvaluator = new MarginEvaluatorWithMarginDecreaseUnoptimizedCnecs(...)
}
if (costOptimization()) {
   ...
} else {
   ...
}


It will also allow us to use the not optimized cnecs in combination with the costly optimization

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/OverloadEvaluator.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Set;
+
+/**
+ * @author Thomas Bouquet {@literal <thomas.bouquet at rte-france.com>}
+ */
+public class OverloadEvaluator implements CostEvaluator {
+    private final Set<FlowCnec> flowCnecs;
+    private final Unit unit;
+    private final MarginEvaluator marginEvaluator;
+    private static final double OVERLOAD_PENALTY = 10000d; // TODO : set this in RAO parameters
```
phiedw, line 26: Did you want to do this in this PR or in a different one?

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/RemedialActionCostEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+        this.rangeActionsOptimizationParameters = rangeActionsOptimizationParameters;
+    }
+
+    @Override
+    public String getName() {
+        return "remedial-action-cost-evaluator";
+    }
+
+    @Override
+    public Pair<Double, List<FlowCnec>> computeCostAndLimitingElements(FlowResult flowResult, RemedialActionActivationResult remedialActionActivationResult, Set<String> contingenciesToExclude) {
```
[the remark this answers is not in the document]
phiedw: I agree it's starting to get a bit confusing. I think we did this to avoid recomputing an ordered list on all cnecs every time we want to have to get the most limiting element(s) for any reason (like logging). But we could still store that list in each relevant evaluator object, and add a getMostConstrainedElements method or something like that to return the wanted info.

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/RemedialActionCostEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+        return Pair.of(getTotalNetworkActionsCost(remedialActionActivationResult) + getTotalRangeActionsCost(remedialActionActivationResult), EvaluatorsUtils.getCostlyElements(flowCnecs, marginEvaluator, unit, flowResult, contingenciesToExclude));
+    }
+
+    private double getTotalNetworkActionsCost(RemedialActionActivationResult remedialActionActivationResult) {
+        return remedialActionActivationResult.getActivatedNetworkActions().stream().mapToDouble(networkAction -> networkAction.getActivationCost().orElse(0.0)).sum();
+    }
+
+    private double getTotalRangeActionsCost(RemedialActionActivationResult remedialActionActivationResult) {
+        double totalRangeActionsCost = 0d;
+        // TODO: this should maybe only be used for MILP equations but not in the evaluator
```
phiedw, line 59: what did you mean with this TODO?

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/RemedialActionCostEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+        for (State state : optimizedStates) {
+            for (RangeAction<?> rangeAction : remedialActionActivationResult.getActivatedRangeActions(state)) {
+                totalRangeActionsCost += rangeAction.getActivationCost().orElse(0d);
+                if (rangeAction instanceof PstRangeAction) {
+                    totalRangeActionsCost += computeVariationCost(rangeAction, rangeActionsOptimizationParameters.getPstPenaltyCost(), state, remedialActionActivationResult);
+                } else if (rangeAction instanceof InjectionRangeAction) {
+                    totalRangeActionsCost += computeVariationCost(rangeAction, rangeActionsOptimizationParameters.getInjectionRaPenaltyCost(), state, remedialActionActivationResult);
+                } else if (rangeAction instanceof HvdcRangeAction) {
+                    totalRangeActionsCost += computeVariationCost(rangeAction, rangeActionsOptimizationParameters.getHvdcPenaltyCost(), state, remedialActionActivationResult);
+                } else {
```
phiedw, line 69: This can be dangerous when we're adding a new type of rangeAction. Make an else if statement for each type of rangeAction and throw in the else

phiedw (changes requested):

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/ObjectiveFunction.java
bqth29: this is for margin optimization where only flow results are needed for evaluation, I figured it was easier to have a version of the method only using a FlowResult as it is used in a lot of parts of the code

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/OverloadEvaluator.java
bqth29: I think I'll do it in a future PR because I don't know when the parameters refactoring will be ready

bqth29: besides it is not a big feature to add

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/RemedialActionCostEvaluator.java
bqth29: i actually updated this in #1194, in the first place I was thinking that we should take in account the range action penalties in evaluators just as we do in the MILP but it is not necessary
I will change it here too

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/RemedialActionCostEvaluator.java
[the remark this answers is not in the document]
bqth29: I agree that we should at least split the method in two for a better clarity

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/LoopFlowViolationCostEvaluator.java
```diff
@@ -43,46 +50,18 @@ public String getName() {
...
     @Override
-    public Pair<Double, List<FlowCnec>> computeCostAndLimitingElements(FlowResult flowResult, Set<String> contingenciesToExclude) {
-        List<FlowCnec> costlyElements = getCostlyElements(flowResult, contingenciesToExclude);
-        double cost = costlyElements
-            .stream()
-            .filter(cnec -> cnec.getState().getContingency().isEmpty() || !contingenciesToExclude.contains(cnec.getState().getContingency().get().getId()))
-            .mapToDouble(cnec -> getLoopFlowExcess(flowResult, cnec) * loopFlowViolationCost)
-            .sum();
+    public CostEvaluatorResult evaluate(FlowResult flowResult, RemedialActionActivationResult remedialActionActivationResult) {
+        Map<FlowCnec, Double> costPerLoopFlowCnec = loopflowCnecs.stream().collect(Collectors.toMap(Function.identity(), loopFlowCnec -> getLoopFlowExcess(flowResult, loopFlowCnec)));
```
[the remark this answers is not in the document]
phiedw: also rename to excess rather than cost

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/LoopFlowViolationCostEvaluator.java
```diff
@@ -43,46 +50,18 @@ public String getName() {
...
+        Map<State, Set<FlowCnec>> flowCnecsPerState = groupFlowCnecsPerState(costPerLoopFlowCnec.keySet());
+        Map<State, Double> costPerState = flowCnecsPerState.keySet().stream().collect(Collectors.toMap(Function.identity(), state -> loopFlowViolationCost * flowCnecsPerState.get(state).stream().mapToDouble(loopFlowCnec -> getLoopFlowExcess(flowResult, loopFlowCnec)).sum()));
```
phiedw, line 56: add line breaks to make this more readable

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/MinMarginEvaluator.java
```diff
@@ -36,65 +43,50 @@ public String getName() {
     }
 
     @Override
-    public Unit getUnit() {
-        return unit;
+    public CostEvaluatorResult evaluate(FlowResult flowResult, RemedialActionActivationResult remedialActionActivationResult) {
+        Set<State> states = flowCnecs.stream().map(FlowCnec::getState).collect(Collectors.toSet());
```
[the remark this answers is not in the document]
phiedw: (with groupFlowCnecsPerState I think?)

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/MnecViolationCostEvaluator.java
```diff
@@ -35,62 +41,27 @@ public MnecViolationCostEvaluator(Set<FlowCnec> flowCnecs, Unit unit, FlowResult
...
-        Collections.reverse(sortedElements);
-
-        return sortedElements;
+    public CostEvaluatorResult evaluate(FlowResult flowResult, RemedialActionActivationResult remedialActionActivationResult) {
+        Map<FlowCnec, Double> costPerMnec = flowCnecs.stream().filter(Cnec::isMonitored).collect(Collectors.toMap(Function.identity(), mnec -> computeMnecCost(flowResult, mnec)));
+        Map<State, Set<FlowCnec>> mnecsPerState = groupFlowCnecsPerState(costPerMnec.keySet());
+        Map<State, Double> costPerState = mnecsPerState.keySet().stream().collect(Collectors.toMap(Function.identity(), state -> Math.abs(mnecViolationCost) < 1e-10 ? 0.0 : mnecViolationCost * mnecsPerState.get(state).stream().mapToDouble(costPerMnec::get).sum()));
+        List<FlowCnec> sortedMnecs = sortFlowCnecsByDecreasingCost(costPerMnec);
+        return new SumCostEvaluatorResult(costPerState, sortedMnecs);
     }
```
phiedw, line 60: use the same structure as for loopflows (just the excess method will look different because of margins vs thresholds)

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/costevaluatorresult/AbsoluteCostEvaluatorResult.java
```diff
@@ -0,0 +1,35 @@
...
+import com.powsybl.openrao.data.cracapi.cnec.FlowCnec;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * @author Thomas Bouquet {@literal <thomas.bouquet at rte-france.com>}
+ */
+public class AbsoluteCostEvaluatorResult implements CostEvaluatorResult {
```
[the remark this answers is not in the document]
bqth29: I used it to indicate that the cost is independent from the contingencies that are filtered out. I also thought of `ConstantCostEvaluatorResult` but figured `AbsoluteCostEvaluatorResult` was better.
But we sure can use another name

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/MinMarginEvaluator.java
```diff
@@ -36,65 +43,50 @@ public String getName() {
...
-        return flowCnecs;
+    private Set<FlowCnec> getFlowCnecsOfState(State state) {
+        return flowCnecs.stream().filter(flowCnec -> state.equals(flowCnec.getState())).filter(FlowCnec::isOptimized).collect(Collectors.toSet());
     }
 
-    @Override
-    public Pair<Double, List<FlowCnec>> computeCostAndLimitingElements(FlowResult flowResult, Set<String> contingenciesToExclude) {
-        List<FlowCnec> costlyElements = getCostlyElements(flowResult, contingenciesToExclude);
+    protected double computeCostForState(FlowResult flowResult, Set<FlowCnec> flowCnecsOfState) {
+        List<FlowCnec> flowCnecsByMargin = flowCnecsOfState.stream().collect(Collectors.toMap(Function.identity(), flowCnec -> marginEvaluator.getMargin(flowResult, flowCnec, unit))).entrySet().stream().sorted(Map.Entry.comparingByValue()).map(Map.Entry::getKey).toList();
```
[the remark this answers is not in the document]
bqth29: indeed it definitely could be used

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/MinMarginViolationEvaluator.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return "min-margin-violation-evaluator";
+    }
+
+    @Override
+    protected double computeCostForState(FlowResult flowResult, Set<FlowCnec> flowCnecsOfState) {
+        return Math.max(0, super.computeCostForState(flowResult, flowCnecsOfState)) * OVERLOAD_PENALTY;
```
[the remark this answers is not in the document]
bqth29: I think a cost per unit is better because if we use a fixed value, then the RAO has no point using more remedial actions to reduce the overload.

For instance, say that you spent 1000€ in remedial actions and there is still a 15 MW overload. We can still use remedial actions but none can secure the network. However, there's one that costs 100€ and that helps reducing the overload to 2 MW.
If we use a constant cost, the virtual cost for overload will be the same, no matter if we apply the 100€-remedial action so the RAO won't choose it.
On the other hand, using a cost per unit (say 10k / unit), we'll go from a cost of 151000 to 21100 so the RAO will chose to apply the last remedial action to reduce the overload.

It all depends on whether we also try to minimize the overload, what are your thoughts on this?

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/MinMarginEvaluator.java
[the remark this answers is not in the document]
bqth29: indeed I forgot to use it here

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/costevaluatorresult/AbstractStateWiseCostEvaluatorResult.java
```diff
@@ -29,28 +29,23 @@ protected AbstractStateWiseCostEvaluatorResult(Map<State, Double> costPerState,
...
     public List<FlowCnec> getCostlyElements(Set<String> contingenciesToExclude) {
-        return costlyElements.stream().filter(flowCnec -> filterFlowCnecOnContingency(flowCnec, contingenciesToExclude)).toList();
+        return costlyElements.stream().filter(flowCnec -> stateHasContingencyToExclude(flowCnec.getState(), contingenciesToExclude)).toList();
     }
 
-    private static boolean filterStateOnContingency(State state, Set<String> contingenciesToExclude) {
+    private static boolean stateHasContingencyToExclude(State state, Set<String> contingenciesToExclude) {
         Optional<Contingency> contingency = state.getContingency();
         return contingency.isEmpty() || !contingenciesToExclude.contains(contingency.get().getId());
     }
```
phiedw, line 50: The name doesnt correspond to what is returned. I would change the return value rather than the naming to avoid double negatives.

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/MinMarginViolationEvaluator.java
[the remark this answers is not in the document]
phiedw: You're right it makes sense to try minimize the excess in the cases where there's not enough linear range actions to solve all overloads for instance.

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/objectivefunctionevaluator/MnecViolationCostEvaluator.java
phiedw: I don't think it's worth making an abstract class for now. But use the same structure with the log. And also the same way of computing the costPerState (I'm not sure why the mnec one requires checking the value of mnecViolationCost but not the loopflow one)

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/costevaluatorresult/AbstractStateWiseCostEvaluatorResult.java
bqth29: I replaced the return value by `contingency.isPresent() && contingenciesToExclude.contains(contingency.get().getId())` but Sonar did not like that the value was always negated. So I went for `statesContingencyMustBeKept` instead.

phiedw (dismissed):

phiedw (approved):

Issue #1084, opened 2024-07-05
Make RaoParameters generic


pet-mit:
### Describe the current behavior

The RaoParameters object (same applies to its json representation) mixes business parameters (such as objective function) with serach-tree-specific parameters (such as maximum search depth).

### Describe the expected behavior

Taking as an example PowSyBl's LoadFlowParameters, separate RaoParameters into:
- "business" RAO parameters that shall apply to any RAO implementation
- extensions containing implementation-specific parameters
 - begin with a SearchTeeRaoParameters extension,which contains parameters specific to SearchTree

### Describe the motivation
This would help create a generic RAO API, and make the two following things possible:
- move the API to powsybl-core
- create new RAO implementations

### Details

RaoParameters:
(B = business parameters, ST = search-tree specific parameter)

| level 1                               | level 2                                                                                                                              | use case | what to do                                                                                                                                                                       |
| ------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------ | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Objective Function Parameters         |                                                                                                                                      |          |                                                                                                                                                                           |
|                                       | Type:  MAX_MIN_MARGIN_IN_MEGAWATT,  MAX_MIN_MARGIN_IN_AMPERE, MAX_MIN_RELATIVE_MARGIN_IN_MEGAWATT, MAX_MIN_RELATIVE_MARGIN_IN_AMPERE | B        | Replace with these generic types : SECURE_FLOW / MAX_MIN_FLOW_MARGIN / MAX_MIN_RELATIVE_FLOW_MARGIN. Unit should be detected depending on LF DC flag **-> obsolete, replaced by: New parameter Unit to store value MEGAWATT or AMPERE**.                     |
|                                       | ForbidCostIncrease                                                                                                                   | ST       | Remove, always set to true in OpenRAO                                                                                                                                     |
|                                       | PreventiveStopCriterion: MIN_OBJECTIVE, SECURE                                                                                       | B+ST     | Remove, compute depending on objective function                                                                                                                           |
|                                       | CurativeStopCriterion: MIN_OBJECTIVE, SECURE, PREVENTIVE_OBJECTIVE, PREVENTIVE_OBJECTIVE_AND_SECURE                                  | B+ST     | Remove, compute depending on objective function                                                                                                                           |
[...]

PR #1197, opened 2024-11-19
Rao parameters separate unit from objective function type

pjeanmarie:

https://github.com/powsybl/powsybl-open-rao/issues/1084

- [x] Yes
- [ ] No

- [x] The *Breaking Change* or *Deprecated* label has been added
- [x] The migration steps are described in the following section

        if "type" in obj_fun:
            if "MAX_MIN_MARGIN" in obj_fun["type"]:
                new_obj_fun["type"] = "MAX_MIN_MARGIN"
            elif "MAX_MIN_RELATIVE_MARGIN" in obj_fun["type"]:
                new_obj_fun["type"] = "MAX_MIN_RELATIVE_MARGIN"
            if "MEGAWATT" in obj_fun["type"]:
                new_obj_fun["unit"] = "MW"
            elif "AMPERE" in obj_fun["type"]:
                new_obj_fun["unit"] = "A"

**Other information**:


commons/src/main/java/com/powsybl/openrao/commons/Unit.java
```diff
@@ -42,6 +42,15 @@ public String toString() {
         return symbol;
     }
 
+    public static Unit getEnum(String value) {
```
obrix, line 45: Necessary ? Unit seems to be serialized properly so why handle case sensitivity when deserializing ?

obrix (commented):
Small comment on deserialization of unit, it may be simplified.

commons/src/main/java/com/powsybl/openrao/commons/Unit.java
pjeanmarie: No reason, I am going to fix it.

phiedw (approved):

PR #1214, opened 2024-12-05
Test building with GraalVM 17

olperr1:
**Other information**:

Test PR  
The CI SNAPSHOT fails for openRAO: unit test fails only on mac!  
The purpose of this test is to check if using a GraalVM-17 JDK instead of the default java17 JDK is the cause of the problem.


olperr1:
The problem is solved by #1234.

PR #1311, opened 2025-03-20
changed critical branch reader

phiedw:

No

Bug fix

When importing cnecs in the cbcora file, we check (for example for temporary limits) tempFactors and fallback to factor if it's not present, and tempA values and fallback to A values if not present.

We now check for tempFactor and tempA, and if both are missing fallback to factor and A values.

- [ ] Yes
- [x] No

**Other information**:


phiedw:
Waiting for info from process before acting

PR #1331, opened 2025-04-02
Move jniortools import into a separate thread.

danielthirion:

- [ ] Yes
- [ ] No

**Other information**:


bqth29 (dismissed):

pet-mit (commented):
I would advise against adding this extra complexity, and instead loading OR-Tools inside the main entry point of the RAO (CastorFullOptimization), even though it's not the best place to do that. I think that avoiding this complexity is a good enough argument to force loading the library in CastorFullOpt (before actually needing OR-Tools). In all cases, OR-Tools is a maven dependency of OpenRAO. 

Would this alternative solution fix the issue you were facing?

danielthirion:
@pet-mit I'll test it after implementing it as a static initializer on CastorFullOpt and I'll let you know.
If the time window between start of computation and loading of library is small enough (a few milliseconds for instance), I believe we can indeed agree that there is no need to put it in a separate thread.

danielthirion:
@pet-mit having tested your proposed solution : 
- moving the static initializer into CastorFullOptimization indeed narrows down the time window where an interruption results in a crash (i.e. time window = time between start of computation and effective library loading) from ~1min (prior to any fix) to ~10s. It roughly is the time it takes for the runner to reach the beginning of the CastorFullOptimization. It is a nice improvement but it might not be enough.
- I therefore moved the loading step into the class **Castor** (which is the class that initializes CastorFullOptimization and is an AutoService). This ensures the library is loaded upon the pod start, instead of waiting for a computation to begin (I confirmed it by putting temporary logs). As far as I am concerned it is the most efficient solution, since it both completely remove the risk of the interruption and does not add any complexity.

What do you think of this solution ?

danielthirion:
> > @pet-mit having tested your proposed solution :
> > 
> > * moving the static initializer into CastorFullOptimization indeed narrows down the time window where an interruption results in a crash (i.e. time window = time between start of computation and effective library loading) from ~1min (prior to any fix) to ~10s. It roughly is the time it takes for the runner to reach the beginning of the CastorFullOptimization. It is a nice improvement but it might not be enough.
> > * I therefore moved the loading step into the class **Castor** (which is the class that initializes CastorFullOptimization and is an AutoService). This ensures the library is loaded upon the pod start, instead of waiting for a computation to begin (I confirmed it by putting temporary logs). As far as I am concerned it is the most efficient solution, since it both completely remove the risk of the interruption and does not add any complexity.
> > 
> > What do you think of this solution ?
> 
> Castor is OK also, as long as you don't go any more up (into RaoProvider)!

Done.
I did not delete the loading in OpenRaoMPSolver as the loading is still theoretically a prerequisite in this class. In case it has already been loaded, nothing is done there. It is not ideal to have the loading at two places, but better safe than sorry if another entrypoint to OpenRaoMPSolver is added. Bear in mind as well that If we decide to remove the loading in OpenRaoMPSolver, we'll have to add it to every test calling it directly (without passing through Castor).

pet-mit (approved):

pet-mit:
DCO is not passing, you can fix it and force push or open another PR from a clean branch

PR #1359, opened 2025-04-29
Refactor network action management for state-based mapping

RoxaneChen02:

This pull request refactors the logic for network action result management to support state-based mapping. The new mapping addresses scenarios where the same `NetworkAction` is activated in multiple states or contingencies (e.g., `contingency1-auto` and `contingency2-auto`). Previously, using only a `Set` to store activated network actions prevented tracking duplicate actions across different states.

This new logic is only used in the `RemedialActionCostEvaluator`so that  we can count and associate `NetworkAction` instances with their respective states, ensuring accurate evaluation in the `RemedialActionCostEvaluator`. Usually not an issue because we compute this cost perimeter per perimeter so we don't have duplicate issue. But we will have to be more careful for **FastRao and MARMOT.** 

Kept the function getActivatedNetworkActions() as it is still used without any issue in the code base. 

- [ ] Yes
- [x ] No

**Other information**:


ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/result/api/NetworkActionsResult.java
```diff
@@ -27,7 +29,9 @@ public interface NetworkActionsResult {
     /**
      * It gathers the {@link NetworkAction} that are activated.
      *
-     * @return The set of activated network actions.
+     * @return The map set of activated network actions.
      */
     Set<NetworkAction> getActivatedNetworkActions();
+
+    Map<State, Set<NetworkAction>> getActivatedNetworkActionsPerState();
```
bqth29, line 36: I think that we should always provide a State as an input of the methods from now on, just like we do for RangeActionActivationResults
what do you think @phiedw

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/result/impl/NetworkActionsResultImpl.java
```diff
@@ -7,29 +7,40 @@
...
     public Set<NetworkAction> getActivatedNetworkActions() {
-        return activatedNetworkActions;
+        // Caution: If the same action is used during two different states,
+        // it will appear only once in the resulting Set because sets do not allow duplicates.
+        return activatedNetworkActionsPerState.values().stream().flatMap(Set::stream).collect(Collectors.toSet());
+    }
+
+    @Override
+    public Map<State, Set<NetworkAction>> getActivatedNetworkActionsPerState() {
+        return activatedNetworkActionsPerState;
```
bqth29, line 44: depending on what we decide to do regarding the previous comment, these methods will need to be adapted

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/marmot/results/GlobalRemedialActionActivationResult.java
```diff
@@ -8,41 +8,52 @@
...
+        Map<State, Set<NetworkAction>> globalNetworkActionsResultPerState = new HashMap<>();
+        preventiveTopologicalActions.getDataPerTimestamp()
+            .forEach((timestamp, networkActionsResult) ->
+                networkActionsResult.getActivatedNetworkActionsPerState()
+                .forEach((state, networkActions) -> {
+                    globalNetworkActionsResultPerState.put(state, networkActions);
+                })
+            );
+
+        this.globalNetworkActionsResultPerState = globalNetworkActionsResultPerState;
```
bqth29, line 40: now that we do this, `NetworkActionImpl::copyWithNewId` is no longer used, can you remove it please?

RoxaneChen02: already done ?

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/result/api/NetworkActionsResult.java
RoxaneChen02: @phiedw said that it wasn't necessary for now as NetworkActionResult aren't as complicated as range actions. Don't know if he changed mind since ahah

bqth29: alright then, we can keep it like this for the moment

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/marmot/results/GlobalRemedialActionActivationResult.java
bqth29: I meant remove the method from class NetworkActionImpl since it is not used elsewhere

bqth29 (approved):

PR #1495, opened 2025-10-13
Accelr changes

phiedw:

no

Improves performance

1) We only delete a variant in contingency scenarios when actions have been applied. That way, via getAvailableNetwork, we can reuse this variant. 
2) In FastRao sensi computations, only define the minimum range actions since we only need their flow values. We could have used a loadflow computation but it was convenient to use our sensi computation api. Since range actions are no longer in the sensi computation, their initial setpoints are no longer available from this sensi computation. Instead, they are fetched from the initial sensi computed via Castor on the filtered RAO (that contains a small number of cnecs).
3) Add a COST_EPSILON on cost evaluation at stop criterion AT_TARGET_OBJECTIVE_VALUE to prevent continuing computations when initial cost is 0 and targetObjectiveValue is also 0.

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/data/crac/crac-api/src/main/java/com/powsybl/openrao/data/crac/api/rangeaction/RangeAction.java
+++ b/data/crac/crac-api/src/main/java/com/powsybl/openrao/data/crac/api/rangeaction/RangeAction.java
@@ -55,7 +55,7 @@ public interface RangeAction<T extends RangeAction<T>> extends RemedialAction<T>
     Optional<String> getGroupId();
 
     /**
-     * Get the marginal cost to spend to increase or decrease the setpoint by one unit.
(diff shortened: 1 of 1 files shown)


data/crac/crac-api/src/main/java/com/powsybl/openrao/data/crac/api/rangeaction/RangeAction.java
```diff
@@ -54,5 +54,20 @@ public interface RangeAction<T extends RangeAction<T>> extends RemedialAction<T>
      */
     Optional<String> getGroupId();
 
+    /**
+     * Get the marginal cost to spend to increase or decrease the setpoint by one unit.
```
Godelaine: variation cost

PR #1526, opened 2025-10-27
feat(Search-Tree): bloomer filter that removes ineffective network action combinations (no impact or duplicate) 

### LINKED ISSUE #1520: `HasImpactOnNetworkFilter` in Search-Tree Bloomer

If all the network actions of a combination have no impact on the network, we still create a leaf an optimize it even though the optimization result won't change.

Filter out the combinations for which ALL the network actions have no impact on network.

### Describe the motivation

_No response_

### Extra Information

_No response_

bqth29:

Fixes #1520 
Copied from: https://github.com/powsybl/powsybl-open-rao/commit/7f0ecfff4f1371bfcad93171e43f733dbe8145db

Network actions is filtered out:
- if it has no impact on network
- if it has the same impact as a less expensive combination

- [ ] Yes
- [ ] No

**Other information**:


ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/searchtree/algorithms/IneffectiveActionsFilter.java
```diff
@@ -0,0 +1,80 @@
...
+
+        return filteredNaCombinations;
+    }
+
+    private double getTotalCostOfCombination(NetworkActionCombination networkActionCombination) {
+        return networkActionCombination.getNetworkActionSet().stream()
+            .mapToDouble(networkAction -> networkAction.getActivationCost().orElse(0.0)).sum();
+    }
+
+    private boolean extractElementaryActionsWithImpact(NetworkActionCombination networkActionCombination, Set<Set<Action>> elementaryActions) {
```
phiedw, line 76: add an explanatory comment to say that elementaryActions is built progressively as we stream the network actions and therefore what it contains when this method is effectively called

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/networkpool/AbstractNetworkPool.java
```diff
@@ -88,6 +88,7 @@ public void releaseUsedNetwork(Network networkToRelease, boolean deleteWorkingVa
 
     protected void cleanVariants(Network networkClone) {
         cleanVariants(networkClone, true);
+        networkClone.getVariantManager().setWorkingVariant(stateSaveVariant);
```
phiedw, line 86: Did you find out why we needed both this line and the one in the SearchTree class?

phiedw (changes requested):

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/networkpool/AbstractNetworkPool.java
bqth29: this line does not seem useful because the tests still work when I comment it
I believe it is more proper to fall back on the initial variant at the end of the pool

bqth29: regarding the line in SearchTree, my guess is that `hasImpactOnNetwork` needs to check values that are specific to variants. At the end of the previous depth, we are probably still on the `workingVariant` of the clone we used for the last computation but it does not exist for the original variant

bqth29: IMO the bug has been here for years, but we just did not see it because none of the filters are relying on variant-specific data

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/searchtree/algorithms/IneffectiveActionsFilter.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Indicate whether a network action combination has an impact on the network. The notion of impact is defined as follows:
+     * <ul>
+     *     <li>at least one of its elementary actions has an impact on the network;</li>
+     *     <li>its total impact on the network, after filtering out the ineffective actions, is not identical to a cheaper combination's.</li>
+     * </ul>
+     * <p>
+     * The {@code elementaryActions} set if filled progressively as we stream the network actions. Thus, it contains the sets of effective
```
phiedw:
```suggestion
     * The {@code elementaryActions} set is filled progressively as we stream the network actions. Thus, it contains the sets of effective
```

phiedw (changes requested):

phiedw (approved):

PR #1540, opened 2025-11-04
[PoC] Robust optimization with Marmot

pet-mit:
This is to showcase how we can inject robustness inside the MIP created by Marmot.  
Mip part is pretty straught-forward, but all the wiring around is pretty complex. I think we should develop another IntertemporalRao implementation next to Marmot, and make it much more simple (without some topo optimization parts).


pet-mit:
replaced by #1543

PR #1598, opened 2026-01-14
fix: ensure consistent and sorted return for function getMonitoredSides 

RoxaneChen02:

This PR ensure that set of side is returned in a specific sorted and immutable structure. So that when we write the flow constraint in MIP we always write in same order, side one then side two


data/crac/crac-impl/src/test/java/com/powsybl/openrao/data/crac/impl/FlowCnecImplTest.java
```diff
@@ -596,4 +596,14 @@ void testIsConnected() {
         network.getGenerator("BBE2AA1 _generator").getTerminal().disconnect();
         assertFalse(cnec3.isConnected(network));
     }
+
+    @Test
+    void tesGetMonitoredSides() {
```
Godelaine: typo, testGetMonitoredSides

PR #1627, opened 2026-02-18
feature(German HVDCs): Implement import mechanism for German HVDCs

vbochetRTE:

Updating FB CRAC importer and VirtualHubs importer in order to handle German HVDCs.

German HVDCs cannot be used.

German HVDCs can be used.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/data/crac/crac-io/crac-io-fb-constraint/src/main/java/com/powsybl/openrao/data/crac/io/fbconstraint/FbConstraintCracCreator.java
+++ b/data/crac/crac-io/crac-io-fb-constraint/src/main/java/com/powsybl/openrao/data/crac/io/fbconstraint/FbConstraintCracCreator.java
@@ -53,7 +53,7 @@ import static com.powsybl.openrao.data.crac.io.commons.ucte.UcteNetworkAnalyzerP
  */
 class FbConstraintCracCreator {
 
-    private static void addFbContraintInstants(Crac crac) {
+    private static void addFbContraintInstants(final Crac crac) {
         crac.newInstant("preventive", InstantKind.PREVENTIVE)
             .newInstant("outage", InstantKind.OUTAGE)
             .newInstant("curative", InstantKind.CURATIVE);


tests/src/test/resources/files/cracCreationParameters/epic14/ccp.json
```diff
@@ -0,0 +1,37 @@
+{
+  "crac-factory" : "CracImplFactory",
+  "default-monitored-line-side" : "monitor-lines-on-both-sides",
+  "ra-usage-limits-per-instant" : [ ],
+  "extensions" : {
+    "FbConstraintCracCreatorParameters" : {
+      "timestamp" : "2026-01-27T17:00:00Z",
+      "ics-cost-up" : 10.0,
```
Godelaine: only include useful parameters, not ics-cost for instance

data/crac/crac-io/crac-io-fb-constraint/src/main/java/com/powsybl/openrao/data/crac/io/fbconstraint/FbConstraintCracCreator.java
```diff
@@ -128,32 +138,62 @@ private void createCnecTimestampFilteringInformation(FlowBasedConstraintDocument
...
+        addPstAndTopoRemedialActionsToCrac(complexVariantReadersGroupedByType.get(Optional.<ActionTypeEnum>empty()), crac, creationContext);
+        addPstAndTopoRemedialActionsToCrac(complexVariantReadersGroupedByType.get(Optional.of(ActionTypeEnum.PST)), crac, creationContext);
+        addPstAndTopoRemedialActionsToCrac(complexVariantReadersGroupedByType.get(Optional.of(ActionTypeEnum.TOPO)), crac, creationContext);
+        addHvdcRemedialActionsToCrac(internalHvdcs, complexVariantReadersGroupedByType.get(Optional.of(ActionTypeEnum.HVDC)), crac, ucteNetworkAnalyzer, creationContext);
+    }
+
+    private static void addPstAndTopoRemedialActionsToCrac(final List<ComplexVariantReader> complexVariantReaders,
+                                                           final Crac crac,
+                                                           final FbConstraintCreationContext creationContext) {
+        if (null == complexVariantReaders) {
```
Godelaine: why this new check ?

data/crac/crac-io/crac-io-fb-constraint/src/main/java/com/powsybl/openrao/data/crac/io/fbconstraint/HvdcLineRemedialActionAdder.java
```diff
@@ -0,0 +1,181 @@
...
+            checkUsageRules();
+        }
+    }
+
+    private void checkUsageRules() {
+        final ActionsSetType fromActionsSetType = fromComplexVariantReader.getComplexVariant().getActionsSet().getFirst();
+        final ActionsSetType toActionsSetType = toComplexVariantReader.getComplexVariant().getActionsSet().getFirst();
+
+        if (fromActionsSetType.isPreventive() != toActionsSetType.isPreventive()) {
+            fromComplexVariantReader.invalidateOnInconsistencyOnState("preventive");
```
Godelaine: Log should be generated once for both readers

vbochetRTE: I'm not sure to understand what is the expected behaviour here...
Should I only invalidate one of the two ComplexVariantReaders and let the other one as is? This does not look clean: both should be invalidated as one can not exist without the other.
Should I invalidate both readers, but put a message for one of them only? Again, I'm not sure invalidating a reader without specifying the reason is a good practice.

Could you please tell me what you had in mind for this point, so that I can implement the solution?

Godelaine: Ah yes you're right I was only talking about the logs but I didn't think of the fact that we need 1 log/ complexVariant, keep it this way sorry !

PR #1631, opened 2026-02-19
[FIX] set systematic sensitivity result status to success if at least one perimeter succeeds

phiedw:

This is more of a temporary fix until we refactor the way statuses are computed after a sensitivity calculation (#1632)

Bug fix

If the only cnec of a perimeter is disconnected (which is not too unlikely to happen with FastRAO), the rao thinks the sensitivity has failed for the perimeter but it should just be a success with a flow of 0 on the given cnec

Now if at least one perimeter has a status of SUCCESS in the OLF result, then the status of the SystematicSensitivityResult will be set to SUCCESS.

- [ ] Yes
- [x] No


sensitivity-analysis/src/main/java/com/powsybl/openrao/sensitivityanalysis/SystematicSensitivityResult.java
Godelaine: update this comment

Godelaine (approved):

PR #1790, opened 2026-07-03
fix: little fix of the default parameter file for cucumber test

RoxaneChen02:

When we load parameter for cucumber test, we load "RaoParameters_default.json" first then if a parameter file is added we change the field that diverge from the default file. But in this default file we have computation in DC and hvdcAcEmulation set to true -> incompatible. We always get a warning at the beginning of each test even if the test is in AC.


Godelaine:
Could you in this PR change all the parameter files with dc set to true and hvdcAcEmuation set to true to dc = true and hvdcAcEmulation = false ?

vbochetRTE (approved):

PR #1801, opened 2026-07-09
Add possibility to have several internal HVDCs, add intermediate object pole

### LINKED ISSUE #1849: Add possibility to have several internal HVDCs, add intermediate object pole

Internal HVDCs are imported from the virtual hubs file without the pole Id information.
The importer takes line/converters of each and puts them in a same list. To handle these in CORE/CE Merging, we need to know which lines/converters belongs to which HVDC/HVDC pole.

Internal HVDCs are now imported from the virtual hubs file with the pole Id information. This information is thus also present in the FlowbasedCracCreationParameters when internal HVDCs are present.

That's because in the future, there will be several entries in the virtualhub configuration, one per internal HVDC.

Also in the merging tool, we will be provided net positions forecasts for the internal HVDCs, and to make the link between these and the configuration, a tag in the NPF will be fed with the value in the tag of the pole (later on they want to call it too but not right now).

### Describe the motivation

_No response_

### Extra Information

_No response_

AntoineLimouzinRTE:

Fixes #1849 

Internal HVDCs are imported from the virtual hubs file without the pole Id information.
The importer takes line/converters of each and puts them in a same list. To handle these in CORE/CE Merging, we need to know which lines/converters belongs to which HVDC/HVDC pole.

Internal HVDCs are now imported from the virtual hubs file with the pole Id information. This information is thus also present in the FlowbasedCracCreationParameters when internal HVDCs are present.

That's because in the future, there will be several <InternalHvdc> entries in the virtualhub configuration, one per internal HVDC.

Also in the merging tool, we will be provided net positions forecasts for the internal HVDCs, and to make the link between these and the configuration, a tag in the NPF will be fed with the value in the <id> tag of the pole (later on they want to call it <eic> too but not right now).

- [ ] Yes
- [x] No

**Other information**:


data/virtual-hubs/virtual-hubs-xml/src/main/java/com/powsybl/openrao/virtualhubs/xml/VirtualHubsConfigurationImporter.java
```diff
@@ -115,15 +115,21 @@ private static void importBorderDirections(final List<com.powsybl.openrao.virtua
...
+        final List<HvdcPole> hvdcPoles = internalHvdc.getPole()
+            .stream()
+            .map(VirtualHubsConfigurationImporter::poleTypeToHvdcPole)
+            .toList();
+
+        configuration.addInternalHvdc(new InternalHvdc(internalHvdc.getEic(), hvdcPoles));
+    }
+
+    private static HvdcPole poleTypeToHvdcPole(final PoleType pole) {
         if (pole == null) {
```
Godelaine, line 132: it looks like there is a change of behavior when you import a null pollType, now your're importing an internalHvdc object with a null hvdcPole, before nothing was imported. Could you add a unit test for that case if that's the behavior you're looking for ?

tests/src/test/resources/files/cracCreationParameters/epic14/ccp.json
```diff
@@ -1,34 +1,58 @@
...
-        }, {
-          "node" : "D4AAA21B",
-          "station" : "Numenor"
-        } ],
-        "lines" : [ {
-          "from" : "D7AAA21A",
-          "to" : "D4AAA21B"
-        } ]
-      } ]
+  "crac-factory": "CracImplFactory",
```
Godelaine, line 2: do we really want second eic = null ? I'd have thought the "equivalent" would have been a second pole object in eic "eic" internal-hvdc with a different pole id, something more like JsonFb..Test example

data/crac/crac-io/crac-io-fb-constraint/src/test/java/com/powsybl/openrao/data/crac/io/fbconstraint/parameters/JsonFbConstraintCracCreationParametersTest.java
```diff
@@ -103,44 +116,49 @@ void exportTest() {
...
+                                                                                              }, {
+                                                                                                "node" : "node 1B",
+                                                                                                "station" : "station B"
+                                                                                              } ],
+                                                                                              "lines" : [ {
+                                                                                                "from" : "node 1A",
+                                                                                                "to" : "node 1B"
+                                                                                              } ]
+                                                                                            }, {
+                                                                                              "id" : "EIC_OF_POLE",
```
Godelaine: are we OK with having poles having the same ID ?

Godelaine (changes requested):

Godelaine:
Also, please update the PR description @AntoineLimouzinRTE to match classical PR descriptions. Either you create an issue describing the fix or you can describe everything directly in the PR description. 
Based on the context you're describing, I don't really understand the need for new HVDC id and pole id fields - what new information are you taking from the igm ? and I don't understand what you mean by "more than one entry in the virtual hub configuration"

data/virtual-hubs/virtual-hubs-xml/src/main/java/com/powsybl/openrao/virtualhubs/xml/VirtualHubsConfigurationImporter.java
AntoineLimouzinRTE: that's not the behavior I was looking for, added a filter above to not have null poles

tests/src/test/resources/files/cracCreationParameters/epic14/ccp.json
AntoineLimouzinRTE: we do not, i missed that

data/crac/crac-io/crac-io-fb-constraint/src/test/java/com/powsybl/openrao/data/crac/io/fbconstraint/parameters/JsonFbConstraintCracCreationParametersTest.java
AntoineLimouzinRTE: Changed that; in real life life it would not

data/virtual-hubs/virtual-hubs-api/src/main/java/com/powsybl/openrao/virtualhubs/InternalHvdc.java
```diff
@@ -8,14 +8,10 @@
...
 
 /**
  * @author Vincent Bochet {@literal <vincent.bochet at rte-france.com>}
  */
-public record InternalHvdc(List<HvdcConverter> converters, List<HvdcLine> lines) {
-    public InternalHvdc(List<HvdcConverter> converters, List<HvdcLine> lines) {
-        this.converters = Objects.requireNonNull(converters, "Virtual hubs configuration does not allow adding null hvdc converters");
-        this.lines = Objects.requireNonNull(lines, "Virtual hubs configuration does not allow adding null hvdc lines");
-    }
+public record InternalHvdc(String eic, List<HvdcPole> poles) {
```
Godelaine, line 16: do you want to requireNonNull eic to force it to be defined ?

data/virtual-hubs/virtual-hubs-api/src/main/java/com/powsybl/openrao/virtualhubs/HvdcPole.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package com.powsybl.openrao.virtualhubs;
+
+import java.util.List;
+import java.util.Objects;
+
+public record HvdcPole(String id, List<HvdcConverter> converters, List<HvdcLine> lines) {
+    public HvdcPole(final String id, final List<HvdcConverter> converters, final List<HvdcLine> lines) {
+        this.id = id;
```
Godelaine: do you want to requireNonNull id to force it to be defined ?

Godelaine:
+ update documentation https://powsybl.readthedocs.io/projects/openrao/en/stable/input-data/crac/creation-parameters.html#flow-based-constraint-specific-parameters

data/virtual-hubs/virtual-hubs-api/src/main/java/com/powsybl/openrao/virtualhubs/HvdcPole.java
AntoineLimouzinRTE: yes, done

data/virtual-hubs/virtual-hubs-api/src/main/java/com/powsybl/openrao/virtualhubs/InternalHvdc.java
AntoineLimouzinRTE: yes, and poles while we're at it

Godelaine (approved):

PR #1814, opened 2026-07-22
chore(hvdc ac emulation): update HVDC range action setpoint if acEmulationDeactivationAction is used

### LINKED ISSUE #1779: feat(HVDC RA): update HVDC range action setpoint if acEmulationDeactivationAction is used

[Reminder of how HVDC are handles right now in the RAO](https://powsybl.readthedocs.io/projects/openrao/en/stable/algorithms/castor/special-features/hvdc.html#hvdc-range-action-optimization-in-preventive-and-curative-instants)

After each load flow is executed on the root leaf of a search tree (i.e. before any network action is applied), the load flow may update the HVDC active power setpoint. This updated setpoint is stored in the network model so that sensitivity computations can still converge if AC emulation is later deactivated.

**However, this load-flow-computed setpoint is not reported in the RAO result.**

This leads to the following problem:

- If the RAO decides to deactivate AC emulation, but does not modify the HVDC setpoint because the optimal value is already equal to the setpoint established by the root load flow, then:
-> the acEmulationDeactivation network action is reported as activated;
-> the HVDC range action is not considered as activated, since from the MIP the setpoint did not change during optimization. 

If an application simply replays the remedial actions contained in the RAO result, it will know that AC emulation must be deactivated, but it will not know which active power setpoint should be applied afterward. The required setpoint is the one computed by the root load flow, but it is currently omitted from the result.

______________________________

This is generally not visible in the preventive search tree because the root load flow == the initial setpoint (that is always reported) see cucumber test 2.2.2.3.13: AC Emulation is activated but not the HVDC range action).

The problem is more likely to be an issue in post-contingency search trees. After preventive actions have been applied, the root load flow may converge to an HVDC setpoint that differs from the initial one. Since this updated setpoint is not reported, the RAO result may no longer contain enough information to faithfully reproduce the network state after deactivating AC emulation.

Before the fix:
-> the acEmulationDeactivation network action is reported as activated;
-> the HVDC range action is not considered as activated

After fix:
-> If an acEmulationDeactivation action is activated
-> set the HVDC range action as activated, activated setpoint = setpoint found by the loadflow
--> Option 1 : always activate HVDC range action, even when the setpoint found by the loadflow is equal to the HVDC's range action's setpoint in the last active perimeter. For instance, if initial setpoint = setpoint1, loadflow in preventive root leaf finds setpoint 1, still activate HVDC range action at setpoint 1.
--> Option 2 : only activate HVDC range action when setpoint differs from last active setpoint. This adds complexity to the code because we need to identify last active perimeter, and to store HVDC's range action's setpoint.
-> we should report the "optimized" setpoint of the HVDC line in the RAO result. "Optimized" setpoint is either the setpoint found by MIP or the setpoint found by loadflow before the MIP

### Questions/investigation to be done

- investigate more on how to activate a range action "manually" 
- if for exemple initialSetpoint = 10, after optimization acEmulation is deactivated and the setpoint is 10 -> do we still serialize 
"intialSetpoint" = 10, "activatedState": {"curative": 10}} ?
- Do we still want to report a setpoint for HVDC range action when it is in acEmulation ? as this setpoint make "no sense".

### Describe the motivation

More coherence in the rao result, a better handling and understanding of the HVDC range actions.

### Extra Information

_No response_

### LINKED ISSUE #1791: chore: write a cucumber test where ac emulation is deactivated in curative and hvdc setpoint != initialSetpoint
Create a test that illustrate this issue:

In curative perimeter after contingency1: 
- After preventive actions have been applied, the root load flow converge to an HVDC setpoint (let's call it setpoint_1) that differs from the initialSetpoint (the setpoint is updated in the network etc.)

Before the fix:
-> the acEmulationDeactivation network action is reported as activated;
-> the HVDC range action is not considered as activated

After fix:
-> If an acEmulationDeactivation action is activated
-> consider the HVDC range action as also activated
-> we should report the setpoint_1

Basically the test 2.2.2.3.13 but in curative

RoxaneChen02:

Fixes #1791 
Fixes #1779

- [ ] Yes
- [ ] No

**Other information**:


ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/linearoptimisation/algorithms/IteratingLinearOptimizer.java
```diff
@@ -143,6 +154,42 @@ private static SensitivityComputer runSensitivityAnalysis(final SensitivityCompu
...
 
+    // If an AcEmulationDeactivation action was used in the leaf, add the associated HVDC range action to the list of activated range actions automatically
+    // even if MIP did not activate the range action
+    private static RangeActionActivationResult updateWithHvdcRangeActionIfNecessary(RangeActionActivationResultImpl currentRangeActionActivationResult,
+                                                                                    IteratingLinearOptimizerInput input) {
+
+        Set<NetworkAction> acEmulationDeactivationActions = input.appliedNetworkActionsInPrimaryState()
+            .getActivatedNetworkActions()
+            .stream()
+            .filter(action -> action.getElementaryActions().stream().allMatch(HvdcAction.class::isInstance))
```
phiedw: You also probably want to check that the HvdcAction is disabling AC emulation mode

phiedw (changes requested):

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/linearoptimisation/algorithms/IteratingLinearOptimizer.java
RoxaneChen02: The only network action that is a HvdcAction is the AcEmulationDeactivation action but maybe not I will check

phiedw (dismissed):

bqth29 (approved):

PR #1818, opened 2026-07-27
feat(RA Usage Limits): ensure that no constraint in curative 1 means that the cumulative behavior starts at curative 2 + ad raUsageLimit coherence checker

### LINKED ISSUE #1840: Cumulative effet of the RA Usage Limit start from the first non null instant

Right now, the cumulative effet concern all the instant of the same kind no matter if the instant is limited or not.

For a given limit (ie. max-ra, max-topo-per-tso etc) and a given TSO, the cumulative effet only starts at the first instant where a limit for the given TSO is defined.
ex. 
curative1: max(rao-per-tso = {"FR":1}
curative2: max(rao-per-tso = {"FR":1, "BE": 2}
curative3: max(rao-per-tso = {"FR":1, "BE": 3}

The action from TSO BE activated in curative 1 do not impact the limit in curative 2 and 3.

### Describe the motivation

If an operator want to only allow actions in curative1 for exemple, this feature can be used

### Extra Information

_No response_

bqth29:
…tive usage limits start at curative 2

- [x] A PR or issue has been opened in all impacted repositories (if any)

Fixes #1840

- No constraint in curative 1 means that the cumulative behavior starts at curative 2. Simply added cucumbers test to check this behavior.

- make max-ra default value to null instead of MAX_VALUE

- check that the limit does not decrease from an instant to the next
- check that no null limit for an instant between two instant with non null limit


bqth29:
The test passes because at the start of C2, the usage limit is reduced from 0 to -3 < 0 so the limit is set to 0 by default. Though this is not really a side effect, a coherence check should be made in case no limit exists in C1 but some exist in C2.

RoxaneChen02:
The solution proposed (no limit in curative1, and 0 in curative2 and curative3) will not work if the RAO go through 2P. See test 15.2.4: Multi-curative - with max-ra-per-tso limits and 2P.

data/crac/crac-api/src/main/java/com/powsybl/openrao/data/crac/api/RaUsageLimitsAdder.java
```diff
@@ -14,7 +14,7 @@
  */
 public interface RaUsageLimitsAdder {
 
-    RaUsageLimitsAdder withMaxRa(int maxRa);
+    RaUsageLimitsAdder withMaxRa(Integer maxRa);
```
bqth29, line 17: I think you can keep `int` because I do not see why someone would call `withMaxRa` with a `null` value

data/crac/crac-api/src/test/java/com/powsybl/openrao/data/crac/api/RaUsageLimitsTest.java
```diff
@@ -38,7 +38,7 @@ class RaUsageLimitsTest {
     @Test
     void testNominalBehavior() {
         // check default values
-        assertEquals(Integer.MAX_VALUE, raUsageLimits.getMaxRa());
+        assertEquals(null, raUsageLimits.getMaxRa());
```
bqth29: use assertIsNull

data/crac/crac-impl/src/main/java/com/powsybl/openrao/data/crac/impl/RaUsageLimitsAdderImpl.java
```diff
@@ -35,7 +35,7 @@ public class RaUsageLimitsAdderImpl implements RaUsageLimitsAdder {
     }
 
     @Override
-    public RaUsageLimitsAdder withMaxRa(int maxRa) {
+    public RaUsageLimitsAdder withMaxRa(Integer maxRa) {
```
bqth29, line 38: same comment here

data/crac/crac-api/src/main/java/com/powsybl/openrao/data/crac/api/RaUsageLimits.java
```diff
@@ -22,19 +22,19 @@
...
-    private int maxRa = DEFAULT_MAX_RA;
+    private Integer maxRa = DEFAULT_MAX_RA;
     private Map<String, Integer> maxTopoPerTso = DEFAULT_MAX_TOPO_PER_TSO;
     private Map<String, Integer> maxPstPerTso = DEFAULT_MAX_PST_PER_TSO;
     private Map<String, Integer> maxRaPerTso = DEFAULT_MAX_RA_PER_TSO;
     private Map<String, Integer> maxElementaryActionsPerTso = DEFAULT_MAX_ELEMENTARY_ACTIONS_PER_TSO;
 
-    public void setMaxRa(int maxRa) {
-        if (maxRa < 0) {
+    public void setMaxRa(Integer maxRa) {
```
bqth29, line 36: I think you can keep `int` for the same reason here

data/crac/crac-api/src/main/java/com/powsybl/openrao/data/crac/api/RaUsageLimits.java
```diff
@@ -76,7 +76,7 @@ public void setMaxElementaryActionsPerTso(Map<String, Integer> maxElementaryActi
         this.maxElementaryActionsPerTso = Objects.isNull(maxElementaryActionsPerTso) ? new HashMap<>() : replaceNegativeValues(maxElementaryActionsPerTso);
     }
 
-    public int getMaxRa() {
+    public Integer getMaxRa() {
```
bqth29, line 79: I find it cleaner to return an `Optional<Integer>`

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/RaoUtil.java
```diff
@@ -64,6 +59,98 @@ public static void initData(final RaoInput raoInput, final RaoParameters raoPara
...
+        }
+    }
+
+    /**
+     * Check that remedial action usage limits are coherent.
+     * We do not allow null values for an instant if a limit was defined for a previous and following instant
+     * ex. if curative2 has no max-ra limit for TSO "FR" but curative1 and curative3 have a limit -> throw an error
+     */
+    public static void checkCurativeRaUsageLimit(Crac crac) {
+        List<String> raUsageLimitToCheck = List.of("maxRaPerTso", "maxTopoPerTso", "maxPstPerTso", "maxElementaryActionsPerTso", "maxRa");
```
bqth29, line 115: maybe create an Enum since you need to iterate through this list here and in the switch case above

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/RaoUtil.java
```diff
@@ -64,6 +59,98 @@ public static void initData(final RaoInput raoInput, final RaoParameters raoPara
...
+        List<Instant> sortedCurativeInstants = crac.getSortedInstants().stream()
+            .filter(Instant::isCurative)
+            .collect(Collectors.toList());
```
bqth29: you can use `crac.getInstants(InstantKind.CURATIVE)`

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/RaoUtil.java
```diff
@@ -64,6 +59,98 @@ public static void initData(final RaoInput raoInput, final RaoParameters raoPara
...
+                return new HashMap<>();
+        }
+    }
+
+    /**
+     * Check that remedial action usage limits are coherent.
+     * We do not allow null values for an instant if a limit was defined for a previous and following instant
+     * ex. if curative2 has no max-ra limit for TSO "FR" but curative1 and curative3 have a limit -> throw an error
+     */
+    public static void checkCurativeRaUsageLimit(Crac crac) {
```
bqth29, line 114: don't you want to check that values are not decreasing too?

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/parameters/RangeActionLimitationParameters.java
```diff
@@ -70,7 +70,7 @@ public Map<String, Integer> getMaxElementaryActionsPerTso(State state) {
         }
     }
 
-    public void setMaxRangeAction(State state, int maxRangeActions) {
+    public void setMaxRangeAction(State state, Integer maxRangeActions) {
```
bqth29, line 73: same here using `int` could be more relevant

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/linearoptimisation/algorithms/fillers/RaUsageLimitsFiller.java
```diff
@@ -99,7 +99,7 @@ Map<State, Set<RangeAction<?>>> getAllRangeActionOfStateToConsider(State state)
         if (state.getInstant().isCurative()) {
             return rangeActions.entrySet().stream()
                 .filter(entry -> entry.getKey().getInstant().isCurative())
-                .filter(entry -> !entry.getKey().getInstant().comesAfter(state.getInstant()))
+                .filter(entry -> !entry.getKey().getInstant().comesAfter(state.getInstant())) // come before or same instant
```
bqth29:
```suggestion
                .filter(entry -> !entry.getKey().getInstant().comesAfter(state.getInstant())) // comes before or same instant
```

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/linearoptimisation/algorithms/fillers/RaUsageLimitsFiller.java
```diff
@@ -276,18 +286,24 @@ private void addMaxElementaryActionsPerTsoConstraint(LinearProblem linearProblem
...
-                        OpenRaoMPVariable totalPstRangeActionTapDownwardVariationVariable = linearProblem.getTotalPstRangeActionTapVariationVariable(
-                            pstRangeAction,
-                            state1,
-                            LinearProblem.VariationDirectionExtension.DOWNWARD
-                        );
-                        maxElementaryActionsConstraint.setCoefficient(totalPstRangeActionTapUpwardVariationVariable, 1);
-                        maxElementaryActionsConstraint.setCoefficient(totalPstRangeActionTapDownwardVariationVariable, 1);
+                        // The total variation of a PST at a previous instant is considered if:
+                        // - the previous instant was under a max-elementary-action-constraint
+                        // - the said PST is considered during the studied instant
```
bqth29:
```suggestion
                        // - said PST is considered during the studied instant
```

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/RaoUtil.java
RoxaneChen02: thought about doing that but it is kind of "code heavy" and duplicate the behavior in RaUsageLimits. If we really want to do something clean we should create a RaUsageLimit interface and all the other limit would define the method .getLimit() etc but again a little out of scope for this PR

ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/commons/RaoUtil.java
RoxaneChen02: yes I forgot about that

data/crac/crac-api/src/main/java/com/powsybl/openrao/data/crac/api/RaUsageLimitsAdder.java
RoxaneChen02: since the default value of max-ra is null now, if you define another limit in the crac a RaUsageLimits is still created but the max-ra field will be initialize to null.

phiedw (approved):

